PR #35, opened 2016-09-29
datastore_mongo: set updated_ts on UpsertAttributes

mchalski:
also, selectively set created_ts.

Issues: MEN-649

@maciejmrowiec @kjaskiewiczz @bboozzoo @marekswiecznik

### DIFF
--- a/datastore_mongo.go
+++ b/datastore_mongo.go
@@ -20,6 +20,7 @@ import (
 	"gopkg.in/mgo.v2"
 	"gopkg.in/mgo.v2/bson"
 	"sync"
+	"time"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


datastore_mongo_test.go
```diff
@@ -634,6 +645,14 @@ func TestMongoUpsertAttributes(t *testing.T) {
 			t.Errorf("attributes mismatch, have: %v\nwant: %v", dev.Attributes, tc.outAttrs)
 		}
 
+		//check timestamp validity
+		//note that mongo stores time with lower precision- custom comparison
+		assert.Equal(t, createdTs.Unix(), dev.CreatedTs.Unix())
```
bboozzoo: I used `assert.WithinDuration()` with 1s delta to workaround precision issues. You should be able to replace both asserts with that. Make sure to leave the comment about precision loss.

datastore_mongo.go
```diff
@@ -156,15 +157,20 @@ func (db *DataStoreMongo) UpsertAttributes(id DeviceID, attrs DeviceAttributes)
 	c := s.DB(DbName).C(DbDevicesColl)
 
 	update := makeAttrUpsert(attrs)
-	update = bson.M{"$set": update}
+
+	//set update time and optionally created time
+	now := time.Now()
+	update["updated_ts"] = now
+	update = bson.M{"$set": update,
+		"$setOnInsert": bson.M{"created_ts": now}}
```
bboozzoo: mongo magic ;)

mchalski: indeed, what can you do :)

datastore_mongo_test.go
mchalski: ...meh, but it won't work in Condition, will still have to keep unix times, or am I missing something

bboozzoo: fair enough

PR #39, opened 2016-10-06
docs: update api specs according to guildelines

mchalski:
Issues: MEN-655

Changelog (besides style/layout)
- reduced the search/sort mini-language description - we implement just a minimal subset
- moved lots of stuff from global description to param/struct descriptions

Possible quirks:
- can't see a way to properly document search, except in the description (search params are free-form, attr names are not known a priori
- attribute value is still an 'object' - I couldn't find any way to document multi-type fields properly

This spec is a lot bigger, with lots of examples, so please take a good look.

@michaelatmender @maciejmrowiec

### DIFF
--- a/docs/integrations_api.yml
+++ b/docs/integrations_api.yml
@@ -1,28 +1,26 @@
 swagger: '2.0'
 info:
-  version: "0.1"
-  title: Mender Device Inventory API
+  version: '0.1'
+  title: Device inventory
   description: |
-    The Inventory API allows devices to upload almost arbitrary attributes (software/hardware info, health checks, metrics, etc.) of various data types to the backend. This data can be used by the administrator to monitor the state of registered devices.
+    An API for device attribute management and device grouping. Intended for use by the web GUI.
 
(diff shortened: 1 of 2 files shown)


docs/integrations_api.yml
```diff
@@ -236,24 +265,19 @@ paths:
 
   /devices/{id}/attributes:
```
maciejmrowiec: This functionality is not implemented yet, let's drop it from doc and move it to the ticket so we have parts of documentation to reuse. https://tracker.mender.io/browse/MEN-638

docs/integrations_api.yml
```diff
@@ -141,80 +143,107 @@ paths:
...
           schema:
             $ref: "#/definitions/Error"
         500:
-          description: Internal error
+          description: Internal server error.
           schema:
             $ref: "#/definitions/Error"
     put:
-      summary: Add device to group.
+      summary: Add a device to a group
```
maciejmrowiec: Important property is that device can be only in one group. Adding it to another will move it to the new group.

maciejmrowiec:
thanks

PR #44, opened 2016-10-28
docs: change attribute 'value' field type to 'string'

mchalski:
helpful for tests - swagger python client doesn't handle 'object' well

@GregorioDiStefano @maciejmrowiec

### DIFF
--- a/docs/devices_api.yml
+++ b/docs/devices_api.yml
@@ -82,7 +82,7 @@ definitions:
         type: string
         description: Attribute description.
       value:
-        type: object
+        type: string
         description: |
             The current value of the attribute.
 
--- a/docs/integrations_api.yml
+++ b/docs/integrations_api.yml
@@ -356,7 +356,7 @@ definitions:
         type: string
         description: Attribute description.
       value:
-        type: object
+        type: string
         description: |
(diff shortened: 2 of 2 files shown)


GregorioDiStefano:
if I recall correctly, this was originally set to object, to indicate that any value, can be used, correct?

mchalski:
@GregorioDiStefano yup

docs/devices_api.yml
```diff
@@ -82,7 +82,7 @@ definitions:
         type: string
         description: Attribute description.
       value:
-        type: object
+        type: string
```
maciejmrowiec: Isn't it a bit misleading? it can be many different types.
I know of few more places we will need to use 'object' without specific structure as we are going to display metadata of artifacts which are allowed to have any structure defined by the user.

mchalski: both are, but @GregorioDiStefano was having problems with 'object' and asked me to modify it.

feel free to discard this PR, or accept it as you see fit.

maciejmrowiec: let's stick you your fix, talked to @GregorioDiStefano

PR #45, opened 2016-10-31
limit deviceid to 1024 bytes, since this limit is imposed by mongo

GregorioDiStefano:


GregorioDiStefano:
@maciejmrowiec please take a look

PR #53, opened 2017-01-10
relative urls in Location and Link headers

mchalski:
Issues: MEN-881

https://tools.ietf.org/html/rfc3986#section-5.4


mchalski:
@bboozzoo fixed

PR #59, opened 2017-01-30
Mongo schema version

mchalski:
insert the service's current version into the db using the dummy migrator from `go-lib-micro/mongo/migrate`.

Issues: MEN-889

@maciejmrowiec @mendersoftware/rndity if there are no issues I'm going to do the same for all other services


main.go
```diff
@@ -58,5 +58,15 @@ func main() {
 	l.Printf("Inventory Service, version %s starting up",
 		CreateVersionString())
 
+	db, err := NewDataStoreMongo(config.Config.GetString(SettingDb))
+	if err != nil {
+		l.Fatal("failed to connect to db")
+	}
+
+	err = db.Migrate(Version, nil)
```
maciejmrowiec: Versioning should be synced with application version otherwise it will be tricky to keep track of this. I would suggest to use stuff from `version.go` as it's autoset by build system so when we set tags or branch in git, mongo version would moved automatically.

mchalski: I don't want this studd to change automatically, and contain git revisions. Why would we need this?

maciejmrowiec: I guess only catch would be on interpreting non tagged builds (tags: 1.0.1, 1.2.4 etc) while branches would go into 1.0.x etc

maciejmrowiec: We don't want to maintain version string manually, what if we forget to change a string and release?

mchalski: but this is the *schema* version. it goes hand in hand with the app, but only as far as it's useful for us. no reason to mix `version.go` (Commit, Tag, Branch, BuildNumber...?) into this.

 I still fail to see any connection whatsowever. The schema version changes when we say it does.

bboozzoo: Actually I would prefer to maintain DB schema version manually. Using git revision seems like an overkill. Unless your goal is to have random looking versions IDs instead of numbers. But then, it's rather complicated to figure our how these versions are ordered. On top of this, I don't see how we're going to organize the process of assigning IDs in such case. You don't know the ID. So even if we take n-1 commit, if you rebase, the hash would change, hence you'd have to update schema version.

Besides, version information in `version.go` is only set when building in travis. I don't think that anyone sets it when building locally. Just another source of confusion.

maciejmrowiec: Hmm but it's tied with version in URI that's API/App version - therefore I made assumption this is application version and that we update that with every single release (in that case it has to be automated). If you are thinking about having separate versioning for db (can be even simple uint ) this we bump only when we introduce some kind of migration then it's OK. Maybe we won't be able to use it also for enforcing max +1 minor upgrade, but for the main thing it would do just fine.

mchalski: it was just never in the proposal, you prepare a migration when you have something to do with the schema only. all of this is done manually and carefully, no magic.

mchalski: @maciejmrowiec no, we don't have to do it every release. we only do it when an actual migration is  needed, until then everything stays at the initial version (hence the jira comment about gaps in schema versions - they're ok). 

so there's no way to forget anything, it'll be a lot of targeted manual work.

mchalski: @maciejmrowiec you're right about a separate db version though. I took a shortcut and reused the api version, that's confusing.

maciejmrowiec: got it, then I guess it's just a misunderstanding from reusing Version var from URI.

maciejmrowiec:
@mchalski I removed `codecov/patch` from required, looks like forgot to do this repo when was cleaning this up

mchalski:
@maciejmrowiec thx, merging

PR #72, opened 2017-02-28
Remove device

mchalski:
Issues: MEN-986

### DIFF
--- a/api_inventory_test.go
+++ b/api_inventory_test.go
@@ -1079,3 +1079,58 @@ func TestApiGetDeviceGroup(t *testing.T) {
 		runTestRequest(t, apih, tc.inReq, tc.JSONResponseParams)
 	}
 }
+
+func TestApiDeleteDevice(t *testing.T) {
+	t.Parallel()
+	rest.ErrorFieldName = "error"
+
+	tcases := map[string]struct {
+		utils.JSONResponseParams
+
+		inReq        *http.Request
+		inDevId      DeviceID
+		inventoryErr error
+	}{
+		"no device": {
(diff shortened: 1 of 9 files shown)


api_inventory_test.go
```diff
@@ -1079,3 +1079,54 @@ func TestApiGetDeviceGroup(t *testing.T) {
...
+			JSONResponseParams: utils.JSONResponseParams{
+				OutputStatus:     http.StatusInternalServerError,
+				OutputBodyObject: RestError("internal error"),
+			},
+			inventoryErr: errors.New("inventory internal error"),
+		},
+	}
+
+	for name, tc := range tcases {
+		t.Logf("test case: %s", name)
```
maciejmrowiec: instead of logf use subtests, wrap everyting under for with:

t.Run(name, func(b *testing.B) {
    t.Parallel()

    inv := MockInventoryApp{}
    ....
    runTestRequest(t, apih, tc.inReq, tc.JSONResponseParams)
}

maciejmrowiec: t.Parallel will allow to run cases concurrently 

https://blog.golang.org/subtests

maciejmrowiec:
otherwise looks good, test comments are just suggestions

api_inventory_test.go
bboozzoo: Just a reminder, you'll have to use a pattern like this: https://github.com/mendersoftware/deviceauth/blob/master/api_devauth_test.go#L472 to be able to use `t.Parallel()` with `for` loop. Otherwise `tc` is just a loop variable and all test cases will run with the same (last) `tc`

PR #76, opened 2017-03-24
calculate code coverage when running acceptance tests

GregorioDiStefano:


GregorioDiStefano:
testing..

GregorioDiStefano:
@kacf @maciejmrowiec 

This wasn't as easy as it looks to get right, but it's working now.
You have to first exit the binary (sending SIGINT signal) and copy the file to the host machine and share it with codecov.io

GregorioDiStefano:
@bboozzoo

.travis.yml
```diff
@@ -122,27 +124,35 @@ script:
     # Copy script to testing directory
     - cp integration_new/extra/travis-testing/* tests/
 
-    # Build docker image from docker file
-    - sudo docker build -t $DOCKER_REPOSITORY:pr .
-
-
```
kacf: Careful with this rename. The deploy section unconditionally pushes the `pr` tag, so this will either cause that step to fail, or worse, cause it to push an old, unrelated, and possibly broken `pr` tag. Maybe you should check the job type in the deploy section as well.

main_test.go
```diff
@@ -0,0 +1,41 @@
...
+	if !runAcceptanceTests {
+		t.Skip()
+	}
+
+	go main()
+
+	stopChan := make(chan os.Signal)
+	signal.Notify(stopChan, os.Interrupt)
+
+	<-stopChan
```
kacf: Clever! :-)

.travis.yml
GregorioDiStefano: Damn, nice find.

Dockerfile.acceptance-testing
```diff
@@ -0,0 +1,8 @@
+FROM alpine:3.4
+
+EXPOSE 8080
+
+COPY ./inventory /usr/bin/
+COPY ./config.yaml /usr/bin/
+
```
bboozzoo: you can add `STOPSIGNAL SIGINT` instead of `docker kill -s SIGINT..`

Also since we're mounting TESTS_DIR into the container at `/tests` you can dump coverprofile to `/tests/coverage-acceptance.txt` and get rid of `docker cp ..`

bboozzoo: BTW. i'm wondering, if we already tag an image as mendersoftware/inventory:pr you could use that image as `FROM`

GregorioDiStefano:
I don't think there is anything left to do here

PR #77, opened 2017-03-28
api: http: do replace 404 with 204 on DELETE if device was not found.

bboozzoo:
Do not return 404 if DELETE on /devices/{id} did not find any resources to
delete. In the end, the effect of DELETE if there was no device and if one was
removed is exactly the same (device is no longer present), but non-success
status may confuse some integration tools.

@maciejmrowiec @mendersoftware/rndity


bboozzoo:
BTW. the tests fail under Go 1.8 due to some changes in format of `json.Unmarshal()` errors.

PR #80, opened 2017-04-04
ignore testing packages when calculating code coverage

GregorioDiStefano:
Changelog: none
Signed-off-by: Greg Di Stefano <greg.distefano+github@gmail.com>

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -1,4 +1,7 @@
 #!/bin/bash
 cd testing
-cp ./docs/management_api.yml .
+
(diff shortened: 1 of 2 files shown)


tests/run.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 cd testing
-cp ./docs/management_api.yml .
+sleep 5
```
bboozzoo: can you try with something like this? https://github.com/mendersoftware/deviceauth/blob/master/tests/run.sh#L11

PR #98, opened 2017-10-04
tests: create fewer groups in acceptance tests

bboozzoo:
When creating a large number of groups within a short period of time will cause
heavy logging. This may in turn cause docker shim to get stuck pushing the logs
from the container outside. A workaround is to either manually `cat` on the
stdout file descriptor of docker shim or reduce logging.

@mendersoftware/rndity @GregorioDiStefano


GregorioDiStefano:
The integration tests actually passed here ^ (s3cmd failed on the new build slave in our USDC)

PR #101, opened 2017-10-10
MEN-811: Get all devices in a group with a single api-call.

oleorhagen:
api/devices?group=<name>

Changelog: Title

Signed-off-by: Ole Petter <ole.orhagen@cfengine.com>


inv/inventory_test.go
```diff
@@ -75,6 +76,13 @@ func TestInventoryListDevices(t *testing.T) {
 			outError:        errors.New("failed to fetch devices: db connection failed"),
 			outDevices:      nil,
 		},
+		"get devices from group": {
```
oleorhagen: I am unsure as to the validity of this test. It passes, but how is it actually working, with the whole mocking functionality and all?

inv/mocks/InventoryApp.go
```diff
@@ -97,22 +97,22 @@ func (_m *InventoryApp) GetDeviceGroup(ctx context.Context, id model.DeviceID) (
 	return r0, r1
 }
 
-// ListDevices provides a mock function with given fields: ctx, skip, limit, filters, sort, hasGroup
-func (_m *InventoryApp) ListDevices(ctx context.Context, skip int, limit int, filters []store.Filter, sort *store.Sort, hasGroup *bool) ([]model.Device, error) {
-	ret := _m.Called(ctx, skip, limit, filters, sort, hasGroup)
+// ListDevices provides a mock function with given fields: ctx, skip, limit, filters, sort, hasGroup, groupName
+func (_m *InventoryApp) ListDevices(ctx context.Context, skip int, limit int, filters []store.Filter, sort *store.Sort, hasGroup *bool, groupName string) ([]model.Device, error) {
```
oleorhagen: Same as above. This has been autogenerated through vektra/mockery.

store/mongo/datastore_mongo.go
```diff
@@ -167,6 +167,10 @@ func (db *DataStoreMongo) GetDevices(ctx context.Context, skip int, limit int, f
 		}
 	}
 
+	if groupName != "" {
+		queryFilters = append(queryFilters, bson.M{DbDevGroup: groupName}) // TODO - is this the right db-call?
```
oleorhagen: I think this is the right db call. It works, but it has not been tested along with filters etc. But I do not believe that should cause a problem?

api/http/api_inventory.go
```diff
@@ -24,12 +24,13 @@ import (
 	u "github.com/mendersoftware/go-lib-micro/rest_utils"
 	"github.com/pkg/errors"
 
+	"regexp"
```
bboozzoo: Move this to the block with stdlib imports. We try to use the following imports order: stdlib, 3rd party, repo local

api/http/api_inventory.go
```diff
@@ -193,7 +201,7 @@ func (i *InventoryHandlers) GetDevicesHandler(w rest.ResponseWriter, r *rest.Req
 	}
 
 	//get one extra device to see if there's a 'next' page
-	devs, err := i.inventory.ListDevices(ctx, int((page-1)*perPage), int(perPage+1), filters, sort, hasGroup)
+	devs, err := i.inventory.ListDevices(ctx, int((page-1)*perPage), int(perPage+1), filters, sort, hasGroup, groupName)
```
bboozzoo: break the line after `int(perPage+1)`

inv/inventory.go
```diff
@@ -25,7 +25,7 @@ import (
 
 // this inventory service interface
 type InventoryApp interface {
-	ListDevices(ctx context.Context, skip int, limit int, filters []store.Filter, sort *store.Sort, hasGroup *bool) ([]model.Device, error)
+	ListDevices(ctx context.Context, skip int, limit int, filters []store.Filter, sort *store.Sort, hasGroup *bool, groupName string) ([]model.Device, error)
```
bboozzoo: `ListDevices` takes 7 parameters now. Maybe it's worth refactoring this a little. How about you add a `ListQuery` struct, something like:
```
type ListQuery struct {
    skip int
    limit int
    filters []store.Filter
    sort *store.Sort // would it be possible to convert this to store.Sort?
    hasGroup bool // don't know why it's a pointer in ListDevices call, 
                  // maybe double check the logic to verify that nothing funky is done here
    group string
}
```

store/datastore.go
```diff
@@ -30,7 +30,7 @@ var (
 )
 
 type DataStore interface {
-	GetDevices(ctx context.Context, skip int, limit int, filters []Filter, sort *Sort, hasGroup *bool) ([]model.Device, error)
+	GetDevices(ctx context.Context, skip int, limit int, filters []Filter, sort *Sort, hasGroup *bool, groupName string) ([]model.Device, error)
```
bboozzoo: same here, use a query struct

store/mongo/datastore_mongo.go
bboozzoo: What do you mean by 'it has not been tested along with filters'? From the code, `queryFilters` is used with mongo's `$and` operator. I think it's ok.

store/datastore.go
GregorioDiStefano: ops, forgot this one..

GregorioDiStefano:

store/query.go
```diff
@@ -30,3 +30,12 @@ type Sort struct {
 	AttrName  string
 	Ascending bool
 }
+
+type ListQuery struct {
```
maciejmrowiec: confusing name, did not know what it is until looked at implementation
probably something like Filter, QueryFilter is closer to what it does

PR #112, opened 2018-10-17
Allow filter with ":"

yongjhih:
We couldn't filter attributes with ":" correctly.

For mac address "ff:ff:ff:ff:ff:ff" example, we will request with the query string like this: `attr_name1=ff:ff:ff:ff:ff:ff`, but actually the filter will be "ff" not "ff:ff:ff:ff:ff:ff".

### DIFF
--- a/api/http/api_inventory.go
+++ b/api/http/api_inventory.go
@@ -147,19 +147,18 @@ func parseFilterParams(r *rest.Request) ([]store.Filter, error) {
 		}
 		valueStrArray := strings.Split(valueStr, queryParamValueSeparator)
 		filter = store.Filter{AttrName: name}
-		valueIdx := 0
 		if len(valueStrArray) == 2 {
-			valueIdx = 1
 			switch valueStrArray[filterEqOperatorIdx] {
 			case "eq":
 				filter.Operator = store.Eq
 			default:
 				return nil, errors.New("invalid filter operator")
 			}
+			filter.Value = valueStrArray[filterEqOperatorIdx+1]
 		} else {
 			filter.Operator = store.Eq
+			filter.Value = valueStr
 		}
-		filter.Value = valueStrArray[valueIdx]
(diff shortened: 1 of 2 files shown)


maciejmrowiec:
hi @yongjhih thanks for contribution!

We require all commits to be signed-off by the authors, details: https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#sign-your-work

We also generate our change logs from the commit history, for this to work it needs extra line in commit message: https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#changelog-tags

Our unit tests are ensuring those are put in place.

There are also some unit tests that are failing in regards to your change, can you please make sure the tests are passing and ideally provide unit test covering for your changes as well?

yongjhih:
The old commit changed the spec it was allowing filter ":" without an "eq" operator, wasn't throwing an exception instead, that's why it failed on one of the test cases that expects an exception.

I fixed the test cases and added certain test cases for testing filter ":"

mchalski:
hi @yongjhih, thanks for the contribution again (and patience, too). finally got around to doing a thorough review.

regarding the core of this PR:
- it fixes one case, but not the most common one (i.e. `attr=eq:ff:ff:...`, but not the implicit `attr=ff:ff:...`)
- it's an issue because the existence of `eq`, or operators in general, is not documented for now; we only promise `attr=...` to work, and don't want to force folks to use `eq` in special cases
- would you consider doing a fix where both the implicit and explicit case is fixed instead? 

a couple nitpicks - I'd suggest:
- squashing both commits into 1, or separating them properly (the 2nd commit - 'Add test cases...' - also contains major logic change in api_inventory.go); this is for keeping the history clean and ease of browsing
- renaming `TestApiInventoryGetDevicesFilter` to `TestApiParseFilterParams`, to distinguish it from the main API method test

yongjhih:
Thanks for the suggestions, it has been done for days.

PR #114, opened 2019-02-06
get total device count on dev-list request to allow leaner pagination in gui

mzedel:
the additional cost per request is most likely set off by the removed requirement for multiple requests to get all the devices per group, that are currently fired every time another group is selected in the ui...

Changelog: None
Signed-off-by: Manuel Zedel <manuel.zedel@northern.tech>

### DIFF
--- a/api/http/api_inventory.go
+++ b/api/http/api_inventory.go
@@ -204,34 +204,28 @@ func (i *inventoryHandlers) GetDevicesHandler(w rest.ResponseWriter, r *rest.Req
 		return
 	}
 
-	//get one extra device to see if there's a 'next' page
 	ld := store.ListQuery{int((page - 1) * perPage),
-		int(perPage + 1),
+		int(perPage),
 		filters,
 		sort,
 		hasGroup,
 		groupName}
 
-	devs, err := i.inventory.ListDevices(ctx, ld)
+	devs, totalCount, err := i.inventory.ListDevices(ctx, ld)
 
 	if err != nil {
 		u.RestErrWithLogInternal(w, r, l, err)
 		return
 	}
 
-	len := len(devs)
-	hasNext := false
-	if uint64(len) > perPage {
-		hasNext = true
-		len = int(perPage)
-	}
-
+	hasNext := totalCount > int(page * perPage)
 	links := utils.MakePageLinkHdrs(r, page, perPage, hasNext)
-
 	for _, l := range links {
 		w.Header().Add("Link", l)
 	}
-	w.WriteJson(devs[:len])
+	// the response writer will ensure the header name is in Kebab-Pascal-Case
+	w.Header().Add("X-Total-Count", strconv.Itoa(totalCount))
+	w.WriteJson(devs)
 }
 
 func (i *inventoryHandlers) GetDeviceHandler(w rest.ResponseWriter, r *rest.Request) {
@@ -395,7 +389,7 @@ func (i *inventoryHandlers) GetDevicesByGroup(w rest.ResponseWriter, r *rest.Req
 	}
 
 	//get one extra device to see if there's a 'next' page
-	ids, err := i.inventory.ListDevicesByGroup(ctx, model.GroupName(group), int((page-1)*perPage), int(perPage+1))
+	ids, totalCount, err := i.inventory.ListDevicesByGroup(ctx, model.GroupName(group), int((page-1)*perPage), int(perPage))
 	if err != nil {
 		if err == store.ErrGroupNotFound {
 			u.RestErrWithLog(w, r, l, err, http.StatusNotFound)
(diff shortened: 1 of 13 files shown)


mchalski:
@manuel first couple comments re the PR itself:
- it doesn't build, please fix
- similar commits need squashing (fixed license year...fixed another license year etc; same with lint); we typically structure PRs by layer (commit1: db + tests + mocks; commit2: app layer..., commit3: api layer); make PRs possible to read and `git log` digestible

api/http/api_inventory.go
```diff
@@ -416,6 +416,8 @@ func (i *inventoryHandlers) GetDevicesByGroup(w rest.ResponseWriter, r *rest.Req
 	for _, l := range links {
 		w.Header().Add("Link", l)
 	}
+	count, err := i.inventory.GetDeviceCountByGroup(ctx, model.GroupName(group))
```
mchalski: needs some error handling

api/http/api_inventory.go
```diff
@@ -416,6 +416,8 @@ func (i *inventoryHandlers) GetDevicesByGroup(w rest.ResponseWriter, r *rest.Req
...
+	w.Header().Add("GroupCount", strconv.Itoa(count))
```
mchalski: I understand it's a way of working around an API response change but am not a fan. group count could easily be part of a json response, no reason to stick it in a custom header. of course that would be an api version bump, and would have to be planned, but if we want it - let's do it the proper way.

mchalski:
@mzedel left 2 more comments. @kjaskiewiczz please chime in.

api/http/api_inventory.go
mzedel: I agree that it would be nice to have this in the response, but thought that it would be tolerable since we have the other pagination information (the page links to lines above) already in the header.
I've seen both ways in other APIs and have no real preference, but know that without this information we struggle in the frontend. So I would vote to keep this in the header as basically a workaround until we have consistent API responses across the endpoints with pagination information.

kjaskiewiczz: what does "GroupCount" mean here?
does it mean there are _GroupCount_ devices in the given group?
I agree with @mchalski that it would be better to design this first (like create a feature request for the backend or, if you want to do it yourself, create PR with just a doc modification, discuss feature there and then continue with implementation).
If you still want to add a custom header, something like "X-total-count" sounds better to me.

kjaskiewiczz:
you should also document this change (update swagger spec)

api/http/api_inventory.go
mzedel: yes, GroupCount as in that amount of devices in the requested group and using "X-.." for the custom header should have been obvious to me. Then I will transfer this to it's own ticket and we can continue the discussion there

kjaskiewiczz:
also, there should be backend-integration-test created to check this additional functionality, but usually we are creating a separate task for integration tests

mchalski:
@mzedel regarding the core of the implementation, I looked at 2 subjects:
- count performance, and if there is anything smarter we can do for counting
- if we can do better than `first list devices, then count devices`, and maybe fold it into a single query

re performance: 
we're already suboptimal by using `skip()` - that's a linear collection scan, and will get worse with every page; and yes `count()` is just as bad, and people move away from counting in favor of other paging strategies...I don't have an answer, and I guess 'it is what it is' and we have to bite the bullet now, we're not going to redefine the whole api now. this is sth to revisit when we hit it big ;)

re query refactoring:
- we *could* improve slightly here (I think), if you're willing to experiment with skip and limit in the aggregation 
- this would require moving the counting logic to the base `store.GetDevicesByGroup`, replacing the skip + limit with its aggregation counterpart (something like https://stackoverflow.com/questions/49031420/mongodb-aggregate-skip-limit-count-for-pagination, so match + skip + limit + try to work counting in there)
- I can't vouch you'll be successful here, and it's a lot of work - so we can leave it as an improvement and just let this PR go (@kjaskiewiczz what do you think?)

regarding the count header:
I also talked with @kjaskiewiczz, let's agree that:
- the header idea is ok
- `X-total-count` is the name
- when you `GET /devices` with no filter - it's just the count of all devices; with a group, it counts group items
- the doc has to be updated (use an online swagger editor as you go - this will catch mistakes before travis does, we do swagger validation there)

mzedel:
after looking through the mongo aggregation pipeline for a bit I think it should be possible to combine both stages (listing and counting) and get the results in one query (I was hoping for a `yield` like function but think it should be possible using `$let` before the skipping part)
at least the count call should be a lot cheaper for all queries that do not require filtering (using the `$collStats` - this should be accurate enough for all our use current cases), but I'll test this locally to at least get an indication about the viability of the approach...

I'll happily change this within this PR (including the header, swagger and `GetDevicesByGroup` moving - but can also let this go and create a new one. whatever works best for you

mchalski:
@mzedel if you're willing to try let's do it in this PR. 

let's maybe timebox this to - 1 day? - depending on how much time you want to spend. if you decide to call it off we'll stay with this solution.

mzedel:
Sounds like a plan @mchalski! I'll try to get through everything tomorrow.

mzedel:
Thanks for reviewing it! Waiting for @kjaskiewiczz's opinion sounds good for me. Just if it is of interest the pipeline boils down to:

```javascript
[
  {
    '$match': {
        '$and': {
            'group': 'groupName',
            'group': { '$exists': true },
            '$or': {
              // whatever comes from the filter-params
            }
        },
    }
  }, {
    '$facet': {
      'results': [
        {
          '$skip': 0
        }, {
          '$limit': 50
        }
      ], 
      'totalCount': [
        {
          '$count': 'count'
        }
      ]
    }
  }, {
    '$project': {
      'results': 1, 
      'totalCount': {
        '$ifNull': [
          {
            '$arrayElemAt': [
              '$totalCount.count', 0
            ]
          }, 0
        ]
      }
    }
  }
]
```

PR #116, opened 2019-02-27
vendor: update govendor

mchalski:
https://tracker.mender.io/browse/MEN-2201

the problem seems to be gone now.


mchalski:
I'm closing this since the problem in fact persists, see notes in issue.

PR #120, opened 2019-03-12
[POC/DON'T MERGE] regex inventory search

mchalski:
https://tracker.mender.io/browse/MEN-1914

this is a POC preceding the breakdown of this epic.

Aims to assess the naive regex search implementation:
- semantically - can it be done, is the api usable
- performance-wise - isn't the performance catastrophic enough to dump the idea

IMO on both fronts we're ok.
The `$regex/~` operator is prototyped and tested with some acceptance tests and seems to work correctly.

A microbenchmark was created to check at least the basics of performance - see the markdown doc.


lluiscampos:
This is now very old, @mchalski can we close it?

mchalski:
@lluiscampos closing. it will be useful at some point but a closed pr is still available for viewing so let's clean up.

PR #127, opened 2019-05-08
docs: internal endpoint for updating device attributes API spec

kjaskiewiczz:
Issues: https://tracker.mender.io/browse/MEN-2474


docs/internal_api.yml
```diff
@@ -63,6 +63,78 @@ paths:
...
             $ref: '#/definitions/Error'
+
+  /devices/{id}:
+    patch:
+      summary: Upload a set of attributes for a device
+      description: |
+        Saves the provided attribute set for the device.
+        This method has upsert semantics:
+        * the values of existing attributes are overwritten
+        * attributes uploaded for the first time are automatically created
```
mchalski: ...and the device is also created

docs/internal_api.yml
```diff
@@ -63,6 +63,78 @@ paths:
...
+      parameters:
+        - name: X-MEN-Source
+          in: header
+          required: true
+          type: string
+          description: |
+              Contains service name.
+              Inventory will use these information togather with 'X-MEN-Msg-Timestamp' to prevent overwriting attributes with older ones.
```
mchalski: this information

docs/internal_api_v1.yml
```diff
@@ -0,0 +1,144 @@
...
+            description: "MAC address"
+          - name: "ports"
+            value:
+              - "8080"
+              - "8081"
+            description: "Open ports"
+  Attribute:
+    description: Attribute descriptor.
+    type: object
+    required:
```
mchalski: also, scope?

docs/internal_api.yml
kjaskiewiczz: fixed

docs/internal_api.yml
kjaskiewiczz: fixed

docs/internal_api_v1.yml
kjaskiewiczz: I think scope shouldn't be returned by the v1 endpoint

PR #145, opened 2019-12-03
Update docker-compose version to 2.1 for acceptance tests

lluiscampos:
Update docker-compose version to 2.1 for acceptance tests to comply with the rest of the integration compose files.

Also, remove docker-compose-integration.yml

### DIFF
--- a/tests/docker-compose-integration.yml
+++ b/tests/docker-compose-integration.yml
@@ -1,5 +1,0 @@
-version: '2'
-services:
-    mender-inventory:
-            # built/tagged locally and only used for testing
-            image: mendersoftware/inventory:pr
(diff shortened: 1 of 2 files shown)


lluiscampos:
@mchalski @kjaskiewiczz can you confirm that what I said in commit 221b53f is true? If so, can I delete the same file in all other backend repos?

mchalski:
@lluiscampos you're right in general, but it's case by case.

it's a leftover, yes, but I'm not sure if it's 100% not used ever, in any repo, so exercise some caution.

lluiscampos:
> @lluiscampos you're right in general, but it's case by case.
> 
> it's a leftover, yes, but I'm not sure if it's 100% not used ever, in any repo, so exercise some caution.

Thank. I'll pay attention and always run the tests on these cases.

PR #152, opened 2020-02-19
MEN-3097: introduce attribute scope, which defaults to 'inventory'

tranchitella:


tranchitella:
Manually started pipeline (this PR needs https://github.com/mendersoftware/integration/pull/817):
https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/119357109

model/device.go
```diff
@@ -31,6 +35,7 @@ type DeviceAttribute struct {
 	Name        string      `json:"name" bson:",omitempty"`
 	Description *string     `json:"description,omitempty" bson:",omitempty"`
 	Value       interface{} `json:"value" bson:",omitempty"`
+	Scope       string      `json:"scope" bson:",omitempty"`
```
alfrunes: Is it really necessary to have a dedicated scope attribute? Wouldn't it suffice to give the name attribute a path-like structure and interpret each separator as a scope separator?
This way we would also achieve nested scopes too.

api/http/api_inventory.go
```diff
@@ -51,6 +51,7 @@ const (
 	queryParamSort           = "sort"
 	queryParamHasGroup       = "has_group"
 	queryParamValueSeparator = ":"
+	queryParamScopeSeparator = "-"
```
alfrunes: can this be "/" or is that not allowed in field names?

tranchitella:
> Looks OK to me, but are we over-complicating scopes a little too much?

@alfrunes oh well, I'm just executing what described in MEN-3097, mainly cherry-picking things from an old feature branch. I don't believe we are over-complicating scopes, though. The diff is huge mainly because I upgraded the vendored go-lib-micro and this requires some changes here and there in the tests.

What do you mean with over-complicating scopes?

model/device.go
tranchitella: @alfrunes yes, it is needed because we don't want to touch API v1, thus inventory scope should be transparent to API consumers. A new (optional, with default) attribute is fine, changing the name with a prefix is not.

api/http/api_inventory.go
tranchitella: @alfrunes the branch I used as reference was using the dash, I like the slash more, let's give it a try indeed. Forced pushed it, should be totally transparent to API consumers.

model/device.go
alfrunes: I'd suggest not having a default at all - the value is simply is what is specified. If one or more separators are specified the name is parsed into scope(s). 
An exception can be added to the parser, setting a default `inventory` scope, but I can't see why we would want this.
Well... I can see the complexity is about the same as already done here (if not more), so ignore the over-complicating part in my review comment ^ :)

api/http/api_inventory.go
tranchitella: It works indeed, and I like the slash more.

model/device.go
tranchitella: @alfrunes we need the default because we are modifying the API v1, which is already consumed in the real world and we cannot break back-compatibility. @mchalski what's your opinion on this?

mchalski: the last thing Mender needs is nested scopes ;) I hope no acceptance criteria mentions this...?

I prefer this implementation as is. clean name, clean scope, it translates verbatim to api responses, minimal translation... mind you we're already doing funny stuff with the attribute struct (prefixed keys, dropped at marshalling).

so short story - I prefer it as is.

api/http/api_inventory.go
```diff
@@ -51,6 +51,7 @@ const (
 	queryParamSort           = "sort"
 	queryParamHasGroup       = "has_group"
 	queryParamValueSeparator = ":"
+	queryParamScopeSeparator = "/"
```
mchalski: originally it was `:` I think. 

I don't mind `/` but I'm worried about the encoding. I can't get a straight answer out of the RFC whether it's a reserved char or not: https://tools.ietf.org/html/rfc3986#section-3.4.

tranchitella: @mchalski it is not and can be safely used

.gitlab-ci.yml
```diff
@@ -1,3 +1,5 @@
+image: golang:1.13
+
```
lluiscampos: @tranchitella For future reference: this actually will not work as expected.

Here you define the default image, but then the templates are overriding the image on each job. In this case there is not undesired effect as the image used by the templates is actually golang:1.13.

The correct way to do it is as it was before: first include the templates, then override the specific jobs that require a different version (or do nothing if we just want to use the "latest Mender supported version"). See documentation [here](https://github.com/mendersoftware/mendertesting/blob/master/.gitlab-ci-check-golang-unittests.yml).

tranchitella: @lluiscampos thanks for clarifying it, do you want me to launch a PR to remove the useless image override?

lluiscampos: A bit late here, but I assume you got that as I have reworked the templates usage I have cleaned up these. So nothing to do, it was only a heads-up to avoid confusion in the future.

PR #154, opened 2020-02-24
MEN-3099: v2 search end-point spec

tranchitella:
Changelog: none

### DIFF
--- a/docs/management_api_v2.yml
+++ b/docs/management_api_v2.yml
@@ -1,0 +1,262 @@
+swagger: '2.0'
+info:
+  version: '1'
+  title: Device inventory filters and search
+  description: |
+    An API for inventory-based filters management and device search.
+    It is intended for use by the web GUI.
+
+    Devices can upload vendor-specific attributes (software/hardware info, health checks, metrics, etc.) of various data types to the backend as scoped attributes.
+
+    This API enables the user to:
+    * search devices by inventory scoped attribute value
+    * use the results to create and manage device groups for deployment scheduling
+
+basePath: '/api/management/v2/inventory'
+host: 'docker.mender.io'
+schemes:
+  - https
+
+paths:
+  /filters/search:
+    post:
+      summary: Search devices based on inventory attributes
+      description:  |
+        Returns a paged collection of devices and their attributes.
+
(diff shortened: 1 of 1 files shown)


docs/management_api_v2.yml
```diff
@@ -0,0 +1,264 @@
...
+          type: string
+          format: Bearer [token]
+          description: Contains the JWT token issued by the User Administration and Authentication Service.
+        - name: body
+          in: body
+          description: The search and sort parameters of the filter
+          schema:
+            type: object
+            required:
+              - userName
```
mchalski: must be a glitch?

docs/management_api_v2.yml
```diff
@@ -0,0 +1,264 @@
...
+
+            Scope is a string and acts as namespace for the attribute name.
+      attribute:
+        type: string
+        description: |
+            Name of the attribute to be queried for filtering.
+      type:
+        type: string
+        description: Type or operator of the filter predicate.
+        enum: [$eq, $ne]
```
mchalski: we went back and forth with Eystein, but the final decision is to not support $ne for now.

(I know it's not a huge effort, but still).

docs/management_api_v2.yml
```diff
@@ -0,0 +1,264 @@
...
+            The current value of the attribute.
+
+            Attribute type is implicit, inferred from the JSON type.
+
+            Supported types: number, string, array of numbers, array of strings.
+            Mixed arrays are not allowed.
+    example:
+      application/json:
+        name: "cpu_load"
+        scope: "metrics"
```
mchalski: iirc all scopes designed by eysteing are: inventory, identity, system, custom. maybe let's use `inventory`? or something from this list, the user might be confused otherwise

docs/management_api_v2.yml
tranchitella: yep, sorry

docs/management_api_v2.yml
tranchitella: remove $ne

docs/management_api_v2.yml
tranchitella: Ok, fixed.

PR #162, opened 2020-04-16
internal endpoint to set the attributes in scope for a device

merlin-northern:
Internal endpoint to set a status of given devices
--

* **deviceauth** calls _update_device_status_ [[1](https://github.com/mendersoftware/workflows/pull/40)] workflow to propagate information on device status to inventory.
* **inventory** exposes internal API to set status for a list of devices [[2](https://github.com/mendersoftware/inventory-enterprise/pull/24)].
* _/filters/search_ can be use by UI to get list of devices as a single place of information.
* a new scope `identity` [4](https://github.com/mendersoftware/integration/pull/881)

related PRs:
 * [1] https://github.com/mendersoftware/workflows/pull/40
 * [2] https://github.com/mendersoftware/inventory-enterprise/pull/24
 * [3] https://github.com/mendersoftware/deviceauth/pull/324
 * [4] https://github.com/mendersoftware/integration/pull/881

ChangeLog:none

### DIFF
--- a/api/http/api_inventory.go
+++ b/api/http/api_inventory.go
@@ -15,6 +15,7 @@
 package http
 
 import (
+	"context"
 	"net/http"
 	"regexp"
 	"strconv"
@@ -22,6 +23,7 @@ import (
 
 	"github.com/ant0ine/go-json-rest/rest"
 	"github.com/go-ozzo/ozzo-validation"
+	midentity "github.com/mendersoftware/go-lib-micro/identity"
 	"github.com/mendersoftware/go-lib-micro/log"
 	u "github.com/mendersoftware/go-lib-micro/rest_utils"
 	"github.com/pkg/errors"
@@ -42,8 +44,9 @@ const (
 	uriGroups        = "/api/0.1.0/groups"
 	uriGroupsDevices = "/api/0.1.0/groups/:name/devices"
 
-	uriInternalTenants = "/api/internal/v1/inventory/tenants"
-	uriInternalDevices = "/api/internal/v1/inventory/devices"
+	uriInternalTenants       = "/api/internal/v1/inventory/tenants"
+	uriInternalDevices       = "/api/internal/v1/inventory/devices"
+	urlInternalDevicesStatus = "/api/internal/v1/inventory/tenants/:tenant_id/devices/:status"
 )
 
 const (
@@ -94,6 +97,7 @@ func (i *inventoryHandlers) GetApp() (rest.App, error) {
 
 		rest.Post(uriInternalTenants, i.CreateTenantHandler),
 		rest.Post(uriInternalDevices, i.AddDeviceHandler),
+		rest.Post(urlInternalDevicesStatus, i.InternalDevicesStatusHandler),
 	}
 
 	routes = append(routes)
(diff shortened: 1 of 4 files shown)


merlin-northern:
https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/136596505

merlin-northern:
> LGTM
> The mender-qa pipeline seems to be running with an outdated version of [meta-mender](https://github.com/mendersoftware/meta-mender), I started a new one [here](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/136624735).

thanks!

api/http/api_inventory.go
```diff
@@ -42,8 +44,9 @@ const (
 	uriGroups        = "/api/0.1.0/groups"
 	uriGroupsDevices = "/api/0.1.0/groups/:name/devices"
 
-	uriInternalTenants = "/api/internal/v1/inventory/tenants"
-	uriInternalDevices = "/api/internal/v1/inventory/devices"
+	uriInternalTenants       = "/api/internal/v1/inventory/tenants"
+	uriInternalDevices       = "/api/internal/v1/inventory/devices"
+	urlInternalDevicesStatus = "/api/internal/v2/inventory/devices/:tenant_id/:status"
```
tranchitella, line 49: @merlin-northern to be totally honest, I don't like this URL because it is not semantic, reading the URL:
`/api/internal/v2/inventory/devices/1/status`

I read it as "update the status of device 1". I also don't get why we need to increment the version (v2), as we are not changing the old behaviour, just adding a new end-point. Also, how does this work if you don't have a tenant (e.g., in the open source version)? Does this work?

If the API is "per-tenant", I believe `/api/internal/v1/inventory/tenants/:tenant_id/devices/:status` is a better candidate.

api/http/api_inventory_test.go
```diff
@@ -1526,6 +1526,84 @@ func TestUserAdmApiCreateTenant(t *testing.T) {
...
+	for name, tc := range testCases {
+		t.Logf("test case: %s", name)
+		inv := minventory.InventoryApp{}
+
+		ctx := contextMatcher()
+
+		inv.On("UpsertAttributes",
+			ctx,
+			mock.AnythingOfType("model.DeviceID"),
+			mock.AnythingOfType("model.DeviceAttributes"),
```
tranchitella, line 1596: I would rather implement func-based mocks that check the actual value of the deviceID and deviceAttributes you receive from the UpsertAttributes call.

api/http/api_inventory.go
merlin-northern: yes it works without tenant normally.  I checked last week. I have no problem with changing the url and staying with v1 it was incremented because it is the new way: we accept a set of device ids.

api/http/api_inventory_test.go
merlin-northern: my [pleasure](https://github.com/mendersoftware/inventory/pull/162/commits/a71f3299d8788c8abb31838f631a53b0ac321f7b#diff-e00c71bbd3855c0ab6891e3a9d7a0dbbR1634)

api/http/api_inventory.go
merlin-northern: I added a [test](https://github.com/mendersoftware/inventory/pull/162/commits/a71f3299d8788c8abb31838f631a53b0ac321f7b#diff-e00c71bbd3855c0ab6891e3a9d7a0dbbR1567) scenario for that.

merlin-northern: my [pleasure](https://github.com/mendersoftware/inventory/pull/162/commits/a71f3299d8788c8abb31838f631a53b0ac321f7b#diff-d38bb1bb13c34ab0a761f45bb70cd491R49)

merlin-northern:
I also [added](https://github.com/mendersoftware/inventory/pull/162/commits/a71f3299d8788c8abb31838f631a53b0ac321f7b#diff-bf91771dbf8096a9f3adbef1d4339bcdR26) `AttrScopeIdentity const`

mchalski:
I see Fabio's already on it - just a general comment:
so this is ok for what it is, but I was wondering - why not a generic EP for setting any attribute in any scope?

from the original description `internal endpoint to set the attributes in scope for a device` I can see you were almost there ;) at some point it became `Internal endpoint to set a status of given devices`, what happened?

the reason I'm asking is - a generic EP would actually be useful, and I suspect we'll introduce it anyway to collect attributes from various places in the system (current software version etc, the stuff that will go into dashboards when backend actually supports them.

so, just some thoughts.

merlin-northern:
> I see Fabio's already on it - just a general comment:
> so this is ok for what it is, but I was wondering - why not a generic EP for setting any attribute in any scope?
> 
> from the original description `internal endpoint to set the attributes in scope for a device` I can see you were almost there ;) at some point it became `Internal endpoint to set a status of given devices`, what happened?
> 
> the reason I'm asking is - a generic EP would actually be useful, and I suspect we'll introduce it anyway to collect attributes from various places in the system (current software version etc, the stuff that will go into dashboards when backend actually supports them.
> 
> so, just some thoughts.

this is really incredible! the first implementation was exactly this! all what was left was the comment you spotted. I will create a follow up on this. good point. thanks.

PR #166, opened 2020-04-28
Move timestamp and group to attributes under system scope

alfrunes:
This PR first refactors the DeviceAttributes to better reflect the structure from the API, handling the mongo-specific document formatting at parse-time.
The main task of MEN-3460 was only to move the update timestamp to the "identity" scoped attributes, but it felt natural to also include the create timestamp and group as well.


alfrunes:
Backend integration tests needs to be adjusted to these changes as these new attributes comes unexpected to the tests.

store/mongo/datastore_mongo.go
```diff
@@ -42,7 +42,7 @@ import (
 //       That is, s/DbDevGroup/DbDevAttributesGroupName/
 
 const (
-	DbVersion = "0.2.0"
+	DbVersion = "0.3.0"
```
merlin-northern, line 45: do you think it is a change big enough to make 1.0.0? I think it is.

store/mongo/migration_0_3_0.go
```diff
@@ -0,0 +1,96 @@
...
+	collDevs := m.ms.client.Database(databaseName).Collection(DbDevicesColl)
+	upsertOpts := mopts.Update().SetUpsert(true)
+
+	// Add timestamp attributes to the identity scope
+	cursor, err := collDevs.Find(m.ctx, bson.M{})
+	if err != nil {
+		return err
+	}
+	defer cursor.Close(m.ctx)
+	for cursor.Next(m.ctx) {
```
merlin-northern, line 48: ok, I am against running update for each device. this can be done with single UpdateMany query with rename and set. also: I do not think that leaving group field is a good idea, it cant be changed or accessed by means of API calls.

alfrunes: Yeah, I think there's room for improvement here. I'm not sure if what you're suggesting is possible though, I don't believe it's possible for an update $set operation to use other fields of the document as value (at least not with version 3.6).

store/mongo/datastore_mongo.go
alfrunes: I'm not sure, depends on whether we get rid of the previous version of the document (currently overlapping). If the two versions of the documents remain compatible I think that's minor version material (no strong opinion though).

merlin-northern:
> Backend integration tests needs to be adjusted to these changes as these new attributes comes unexpected to the tests.

yeah, I did it like that when I added a scope:

https://github.com/mendersoftware/integration/pull/881

this PR and the one in hte integration have to go together.

model/device.go
```diff
@@ -131,11 +132,11 @@ func (d *Device) UnmarshalBSON(b []byte) error {
 		return err
 	}
 	for _, attr := range d.Attributes {
-		if attr.Scope == "identity" {
+		if attr.Scope == "identity" && attr.Name == "group" {
```
merlin-northern, line 135: why arent we using `AttrScopeIdentity` in here?
also: `AttrGroup = "group"` or something along these lines wouldnt be good?

store/mongo/migration_1_0_0.go
```diff
@@ -32,26 +32,26 @@ type migration_1_0_0 struct {
...
 
 	databaseName := mstore.DbFromContext(m.ctx, DbName)
 	collDevs := m.ms.client.Database(databaseName).Collection(DbDevicesColl)
 
 	// Move timestamps to identity scope.
 	_, err := collDevs.UpdateMany(m.ctx, bson.M{}, bson.M{
 		"$rename": bson.M{
-			"created_ts": identityScope + "-created_ts.value",
-			"updated_ts": identityScope + "-updated_ts.value",
+			"created_ts": metaScope + "-created_ts.value",
```
merlin-northern, line 43: these patterns are repeated really a lot, I thought maybe define them, or provide a helper (choose better name) 
```go
func valueField(string f) string {
	switch f {
	case "created_ts":
		return metaScope + "-" + f + ".value"
	case "updated_ts":
		return metaScope + "-" + f + ".value"
	case "group":
		return identityScope + "-" + f + ".value"
	}
}
```

model/device.go
```diff
@@ -122,6 +122,41 @@ type Device struct {
...
+// internalDevice is only used internally to avoid recursive type-loops for
+// member functions.
+type internalDevice Device
+
+func (d *Device) UnmarshalBSON(b []byte) error {
+	if err := bson.Unmarshal(b, (*internalDevice)(d)); err != nil {
+		return err
+	}
+	for _, attr := range d.Attributes {
+		if attr.Scope == "identity" {
```
merlin-northern: see the other comment, we do have const for that. it probably happens more times.

model/device.go
```diff
@@ -122,6 +122,41 @@ type Device struct {
...
+			switch attr.Name {
+			case "group":
```
merlin-northern: maybe those three attributes names we should also define as consts?

tests/tests/test_api_internal.py
```diff
@@ -13,62 +13,72 @@
...
+        self, internal_client, management_client, clean_db, inventory_attributes,
+    ):
         # insert first device
-        devid = "".join([ format(i, "02x") for i in os.urandom(128)])
-        _, r = internal_client.create_device(device_id = devid,
-                                             attributes=inventory_attributes)
+        devid = "".join([format(i, "02x") for i in os.urandom(128)])
+        _, r = internal_client.create_device(
+            device_id=devid, attributes=inventory_attributes
+        )
```
merlin-northern, line 75: is this black-formatted?

tests/tests/test_api_internal.py
```diff
@@ -13,62 +13,72 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 import os
-from common import internal_client, management_client, mongo, clean_db, inventory_attributes
+from common import (
```
merlin-northern, line 16: ok, it is hard to tell here, since lots of formatting has changed, what exactly was modified here? the API should stay the same.

store/mongo/datastore_mongo.go
```diff
@@ -209,11 +215,11 @@ func (db *DataStoreMongo) GetDevices(ctx context.Context, q store.ListQuery) ([]
 		"$facet": bson.M{
 			"results": []bson.M{
 				sortQuery,
-				bson.M{"$skip": q.Skip},
+				{"$skip": q.Skip},
```
merlin-northern, line 225: is there a reason for that change?

store/mongo/migration_0_2_0_test.go
```diff
@@ -37,86 +42,161 @@ func TestMigration_0_2_0(t *testing.T) {
 	}{
 		"single dev": {
 			inDevs: []interface{}{
-				model.Device{
+				legacyDevice{
```
merlin-northern, line 49: I am having seconds thoughts about that. technically we are using a different structure in tests than in actual production code, right?

alfrunes: This is solely to be able to check that the previous definition of a device - one where the attributes does not have scopes - is correctly migrated to the updated definition. The updated Device can no longer marshal attributes without already putting the scope-name as part of the attribute field, hence the `legacyDevice` definition. (Note that this definition is defined locally and only used for this particular migration test).

model/device.go
alfrunes: I changed it to AttrScopeSystem.

tests/tests/test_api_internal.py
alfrunes: yeah, I have a hook that runs when saving the file. We should probably do a separate PR formatting all the acceptance test sources so we don't get these irrelevant diffs.

tests/tests/test_api_internal.py
alfrunes: The only thing I've changed here is the logic for validating inventory attributes by filtering the returned attributes by scope == "inventory" (L105:106).

store/mongo/datastore_mongo.go
alfrunes: This is just a side-effect of running gofmt with the `-s` flag ("simplify").

model/device.go
alfrunes: :+1:

alfrunes:
Started a new integration pipeline with mendersoftware/integration#904 here: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/143715148

alfrunes:
@merlin-northern could you take a quick look at the 5 latest commits where I address your comments / move attribute scope to "system".

store/mongo/migration_0_3_0.go
```diff
@@ -0,0 +1,96 @@
...
+		var dev model.Device
+		err := cursor.Decode(&dev)
+		if err != nil {
+			return err
+		}
+		dev.Attributes = append(dev.Attributes,
+			model.DeviceAttribute{
+				Name:  "created_ts",
+				Value: dev.CreatedTs,
+				Scope: model.AttrScopeIdentity,
```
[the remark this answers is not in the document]
alfrunes: This was a line I forgot to change in my first attempt to move these attribute to the `system` scope; fortunately my changes to the backend-integration tests picked that up.

model/device.go
```diff
@@ -109,41 +138,67 @@ func (gn GroupName) String() string {
 }
 
 // wrapper for device attributes names and values
-type DeviceAttributes map[string]DeviceAttribute
+type DeviceAttributes []DeviceAttribute
```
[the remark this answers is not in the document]
alfrunes: I've been trying to the best of my abilities to be careful with this model change, and I'm very confident that this isn't breaking anything. All changes to a device's inventory goes through the `store.mongo.UpsertAttributes` function which creates an upsert document from scratch - so the change in application code is very low. As for the unit tests; I created a bson Marshaller interface for the attributes model in order to get the documents right on `Insert` operations. If one or the other document models were incorrect I would expect the unit tests to fail (but fortunately they're not :).
I'll make sure to trace the diff one more time before moving on with this PR for the sake of being cautious.

store/mongo/migration_0_3_0.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+type migration_0_3_0 struct {
+	ms  *DataStoreMongo
+	ctx context.Context
+}
+
+// Up creates timestamp and group attributes under the identity scope.
+// The values reflect the values previously held in the root of the document.
+func (m *migration_0_3_0) Up(from migrate.Version) error {
```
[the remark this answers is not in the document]
alfrunes: I'm in the working of creating an alternative "maintenance mode" of running this migration without shutting down the service, but there needs to be a migration as well to keep everything in sync.

model/device_test.go
```diff
@@ -75,3 +75,133 @@ func TestDeviceAttributesMarshal(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, "[]", string(data))
 }
+
```
merlin-northern, line 79: general comment for the tests: do we have units that check if given attribute landed in correct scope?

model/device.go
```diff
@@ -152,6 +152,9 @@ func (d *Device) UnmarshalBSON(b []byte) error {
 }
 
 func (d Device) MarshalBSON() ([]byte, error) {
+	if err := d.Validate(); err != nil {
+		return nil, err
+	}
```
merlin-northern, line 157: you lied in the commit message. (just nitpicking)

alfrunes:
Started another integration pipeline with the latest additions at: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/145205678

tests/tests/test_api_internal.py
merlin-northern: looks good then

store/mongo/migration_0_2_0_test.go
merlin-northern: ok. I trust you, you see?

merlin-northern:
> @merlin-northern could you take a quick look at the 5 latest commits where I address your comments / move attribute scope to "system".

I did, and answered on slack instead. eh. looks good to me.

PR #174, opened 2020-05-20
FIX: Incorrect unmarshalling datetime from device document

alfrunes:
Only affects the API-facing `updated_ts` attribute for fetching devices.


alfrunes:
> lgtm!
> could we please add unit test for that scenario?

I'll do that in a follow-up PR, I can see that all unit tests avoid checking the timestamp values, which is how this bug slipped.

PR #175, opened 2020-05-22
MEN-3470: Device group management bulk API

alfrunes:
Implementation of the following two endpoints:
 - DELETE /api/management/v1/inventory/groups/{name}/devices
 - PATCH /api/management/v1/inventory/groups/{name}/devices

Both of which accepts a list of device IDs for clearing and updating a bulk of devices' group.


docs/management_api.yml
```diff
@@ -385,6 +388,129 @@ paths:
...
+            type: array
+            items:
+              type: string
+      responses:
+        200:
+          description: Successful response
+          schema:
+            title: UpdatedCount
+            description: |
+              JSON object listing how many devices was updated.
```
tranchitella, line 424: **were** updated

docs/management_api.yml
```diff
@@ -385,6 +388,129 @@ paths:
...
+        500:
+          description: Internal server error.
+          schema:
+            $ref: '#/definitions/Error'
+
+    delete:
+      summary: Clear devices' group
+      description: |
+        Removes a list of devices from their respective groups. This API provides
+        a bulk alternative for DELETE /devices/{id}/group/{name} for managing
```
tranchitella, line 459: alternative **to** DELETE

docs/management_api.yml
```diff
@@ -385,6 +388,129 @@ paths:
...
+            description: |
+              JSON object listing how many devices was updated.
+            type: object
+            required:
+              - updated_count
+            properties:
+              updated_count:
+                type: number
+                description: |
+                  Number of devices' group that was successfully cleared.
```
tranchitella, line 497: Number of devices for which the group was successfully cleared

model/device.go
```diff
@@ -183,6 +185,26 @@ func (gn GroupName) String() string {
...
+	if len(gn) > 1024 {
+		return errors.New(
+			"Group name can at most have 1024 characters",
+		)
+	} else if len(gn) == 0 {
+		return errors.New(
+			"Group name cannot be blank",
+		)
+	} else if ok, _ := regexp.MatchString(
+		"^[A-Za-z0-9_-]*$", string(gn),
```
merlin-northern, line 199: maybe we could compile that regex, since it does not change through-out the life of the service?

alfrunes: Fixed!

model/device.go
```diff
@@ -183,6 +185,26 @@ func (gn GroupName) String() string {
 	return string(gn)
 }
 
+func (gn GroupName) Validate() error {
```
merlin-northern, line 191: I have a vision of validation go-lib-micro module:
```go
 if validation.AlphaNum(gn) { ... }
```
or even `validation.ValidGroupName(gn)`

model/device.go
```diff
@@ -183,6 +185,26 @@ func (gn GroupName) String() string {
...
+	} else if ok, _ := regexp.MatchString(
+		"^[A-Za-z0-9_-]*$", string(gn),
+	); !ok {
+		return errors.New(
+			"Group name can only contain: upper/lowercase " +
+				"alphanum, -(dash), _(underscore)",
+		)
+	}
+	return nil
+}
```
merlin-northern, line 207: and a general mean nitpick: where are the units for model/Validate?

store/datastore.go
```diff
@@ -60,6 +60,20 @@ type DataStore interface {
 	// Updates device group
 	UpdateDeviceGroup(ctx context.Context, devid model.DeviceID, group model.GroupName) error
 
+	// UpdateDevicesGroup updates multiple devices' group, returning number
+	// of matching devices, the number devices that changed group and error,
+	// if any.
+	UpdateDevicesGroup(
+		ctx context.Context, devIDs []model.DeviceID, group model.GroupName,
```
merlin-northern, line 67: is this formatting on purpose? (same below)

store/mongo/datastore_mongo.go
```diff
@@ -460,6 +460,58 @@ func (db *DataStoreMongo) UpdateDeviceGroup(ctx context.Context, devId model.Dev
...
+func (db *DataStoreMongo) UpdateDevicesGroup(
+	ctx context.Context,
+	devIDs []model.DeviceID,
+	group model.GroupName,
+) (int64, int64, error) {
+	database := db.client.Database(mstore.DbFromContext(ctx, DbName))
+	collDevs := database.Collection(DbDevicesColl)
+
+	filter := bson.M{
+		"_id": bson.M{"$in": devIDs},
```
merlin-northern, line 474: we have this:
[DbDevId              = "_id"](https://github.com/mendersoftware/inventory/blob/master/store/mongo/datastore_mongo.go#L45)

store/mongo/datastore_mongo.go
```diff
@@ -460,6 +460,58 @@ func (db *DataStoreMongo) UpdateDeviceGroup(ctx context.Context, devId model.Dev
...
+	database := db.client.Database(mstore.DbFromContext(ctx, DbName))
+	collDevs := database.Collection(DbDevicesColl)
+
+	filter := bson.D{
+		{Key: "_id", Value: bson.M{"$in": deviceIDs}},
+		{Key: DbDevAttributesGroupValue, Value: group},
+	}
+	update := bson.M{
+		"$unset": bson.M{
+			DbDevAttributesGroup: "",
```
merlin-northern, line 506:
```go
DbDevAttributesGroup: 1,
```
right?

model/device.go
alfrunes: The `ozzo-validation` package has an interface called `Validateable` that, if implemented, is called whenever validation is required for a field of the given type. I think that's pretty neat :)

store/mongo/datastore_mongo.go
alfrunes: It doesn't really matter what the value is, according to the docs:

> The specified value in the $unset expression (i.e. "") does not impact the operation.

https://docs.mongodb.com/manual/reference/operator/update/unset/

store/datastore.go
alfrunes: Yeah, I think I've developed some kind of wide-line-allergy. Especially for lines that wraps around my screen.

api/http/api_inventory.go
```diff
@@ -477,6 +471,72 @@ func (i *inventoryHandlers) GetDevicesByGroup(w rest.ResponseWriter, r *rest.Req
...
+			http.StatusBadRequest,
+		)
+		return
+	} else if len(deviceIDs) == 0 {
+		u.RestErrWithLog(w, r, l,
+			errors.New("no device IDs present in payload"),
+			http.StatusBadRequest,
+		)
+		return
+	} else if err = groupName.Validate(); err != nil {
```
merlin-northern, line 526: just nitpick: I would move the validation to line 513, right after the `groupName := ...`, if the group is not valid then there is no point in decoding the payload.

store/mocks/DataStore.go
```diff
@@ -14,10 +14,13 @@
...
-import context "context"
-import mock "github.com/stretchr/testify/mock"
-import model "github.com/mendersoftware/inventory/model"
-import store "github.com/mendersoftware/inventory/store"
+import (
+	"context"
+
+	"github.com/mendersoftware/inventory/model"
+	"github.com/mendersoftware/inventory/store"
+	"github.com/stretchr/testify/mock"
```
merlin-northern, line 22: as long as you are here, I think we order imports in mendersoftware-last order.

store/mongo/datastore_mongo_test.go
```diff
@@ -2695,6 +2696,233 @@ func TestMongoSearchDevices(t *testing.T) {
...
+				Collection(DbDevicesColl)
+			if _, err := collDevs.InsertMany(ctx, deviceSet); err != nil {
+				t.Fatalf("Failed to initialize test context, error: %v", err)
+			}
+			matched, updated, err := store.UpdateDevicesGroup(
+				ctx, testCase.DeviceIDs, testCase.GroupName,
+			)
+
+			assert.Equal(t, testCase.ExpectedMatches, matched)
+			assert.Equal(t, testCase.ExpectedUpdates, updated)
```
merlin-northern, line 2808: can we add a scenario here, to check that in the db all the devices that where outside of `testCaseDeviceIDs` were not changed afer the call to UpdatedevicesGroup?

store/mocks/DataStore.go
```diff
@@ -274,6 +277,34 @@ func (_m *DataStore) UnsetDeviceGroup(ctx context.Context, id model.DeviceID, gr
 	return r0
 }
 
+func (_m *DataStore) UnsetDevicesGroup(
```
merlin-northern, line 280: and where is the unit test for that? (nitpick mode on)

alfrunes: ohh, there is a test in `store/mongo/datastore_mongo_test.go` but I called it `TestClearDevicesGroup` for some reason.

model/device.go
alfrunes: I added a simple unit test that covers all scenarios.

PR #180, opened 2020-06-04
MEN-3637 internal endpoint to set arbitrary attributes.

merlin-northern:
ChangeLog:none


merlin-northern:
goes with https://github.com/mendersoftware/deviceauth/pull/351

PR #185, opened 2020-06-11
Refactor GetDevices to avoid using the aggregation pipeline

tranchitella:
CHangelog: none

### DIFF
--- a/store/mongo/datastore_mongo.go
+++ b/store/mongo/datastore_mongo.go
@@ -147,11 +147,6 @@ func NewDataStoreMongo(config DataStoreMongoConfig) (store.DataStore, error) {
 	return db, nil
 }
 
-type internalDeviceResult struct {
-	Devices    []model.Device `bson:"results"`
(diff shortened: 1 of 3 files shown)


store/mongo/datastore_mongo.go
```diff
@@ -209,54 +209,26 @@ func (db *DataStoreMongo) GetDevices(ctx context.Context, q store.ListQuery) ([]
...
+	}
 	defer cursor.Close(ctx)
 
-	if !cursor.Next(ctx) {
-		return nil, 0, nil
+	devices := []model.Device{}
+	if err = cursor.All(ctx, &devices); err != nil {
+		return nil, -1, errors.Wrap(err, "failed to search devices")
 	}
-	res := internalDeviceResult{}
```
alfrunes, line 255: You can also delete this type definition (`internalDeviceResult`) as it's only used here.

tranchitella: Right, done!

PR #186, opened 2020-06-19
API spec fixes: schema and hostname

lluiscampos:
* [API spec] Consistently use host hosted.mender.io across all repos

### DIFF
--- a/docs/devices_api.yml
+++ b/docs/devices_api.yml
@@ -8,7 +8,7 @@ info:
     Devices can upload vendor-specific attributes (software/hardware info, health checks, metrics, etc.) of various data types to the backend.
 
 basePath: '/api/devices/v1/inventory'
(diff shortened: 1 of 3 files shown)


docs/internal_api.yml
```diff
@@ -7,9 +7,9 @@ info:
     Not exposed via the API Gateway - intended for internal use only.
 
 basePath: '/api/internal/v1/inventory'
-host: 'mender-inventory:8080'
+host: 'hosted.mender.io'
```
tranchitella, line 10: This API is not exposed on hosted mender, and it is considered internal.
I don'w know what the policy is for these end-points, but hosted.mender.io is wrong.

lluiscampos: Thanks. Fixed now.

PR #191, opened 2020-07-09
MEN-3793 remove device on reject

merlin-northern:
ChangeLog:none

### DIFF
--- a/api/http/api_inventory.go
+++ b/api/http/api_inventory.go
@@ -693,6 +693,10 @@ func (i *inventoryHandlers) InternalDevicesStatusHandler(w rest.ResponseWriter, 
 	}
 
 	for _, id := range ids {
+		if status == "rejected" {
+			i.inventory.DeleteDevice(ctx, model.DeviceID(id))
+			continue
+		}
 		attrs := model.DeviceAttributes{
 			{
 				Name:        "status",


api/http/api_inventory.go
```diff
@@ -693,6 +693,10 @@ func (i *inventoryHandlers) InternalDevicesStatusHandler(w rest.ResponseWriter,
 	}
 
 	for _, id := range ids {
+		if status == "rejected" {
```
alfrunes, line 696: Can you make this such that only one mongodb "DeleteMany" operation is executed for this case?
Otherwise I'm all for this change. :+1:

merlin-northern: ok, deviceauth is the only one who calls it, and always with a single id. the only place that calls it with multipe ids is migration command, but it was run only once so far.

merlin-northern:
> Since the endpoint can take an array of device ids it's good practice to make the calls to mongo as efficient as possible (this does not only apply to the "rejected" status but also to Update opeartions of other statuses). But since this is a "hotfix" I'll approve it for now and suggest it to be followed up later.

thanks a lot! I will create the follow up!

merlin-northern:
> > Since the endpoint can take an array of device ids it's good practice to make the calls to mongo as efficient as possible (this does not only apply to the "rejected" status but also to Update opeartions of other statuses). But since this is a "hotfix" I'll approve it for now and suggest it to be followed up later.
> 
> thanks a lot! I will create the follow up!

https://tracker.mender.io/browse/MEN-3811

PR #193, opened 2020-07-29
MEN-3811: Add bulk interface for modifying and deleting devices

alfrunes:
I also removed former single-document interfaces from the storage layer, performing the abstraction in the app layer instead in order to avoid duplicating code and tests.
- DeleteDevice -> DeleteDevices
- UpsertAttributes -> UpsertDevicesAttributes
- UnsetDeviceGroup -> UnsetDevicesGroup
- UpdateDeviceGroup -> UpdateDevicesGroup


tranchitella:
@alfrunes can you please have a look at the unit tests that are failing?

api/http/api_inventory.go
```diff
@@ -772,50 +772,60 @@ func getTenantContext(ctx context.Context, tenantId string) context.Context {
...
-			err = i.inventory.DeleteDevice(ctx, model.DeviceID(id))
-		} else {
-			//upsert the attributes
-			err = i.inventory.UpsertAttributes(ctx, model.DeviceID(id), attrs)
-		}
-		if err != nil {
-			u.RestErrWithLogInternal(w, r, l, err)
-			return
-		}
+	case StatusDecommissioned, StatusRejected:
```
mchalski, line 811: where did the `StatusRejected` come from?

`decommissioned` is a very special case (rejecting preauth'd authsets) and it's one case when we want to get rid of the device.

we don't want the get rid of all device data if it's rejected.

mchalski: FYI there's a related discussion going on under a bug I'm working on:
https://tracker.mender.io/browse/MEN-2558?focusedCommentId=108637&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-108637

note that we stick to the philosophy of not removing devices even if there are no authsets; 
we'll possibly go as far as introducing a new `noauth` state just to keep a device around.

that decision was made a long long time ago, but Eystein just reiterated it.

alfrunes: Thanks for sharing, I was not aware of this and understood that we should only keep the rejected entry in deviceauth (after all, the device has no way of updating the inventory once it's rejected). I'll move the rejected status to the previous case then.

mchalski: thank you! yes, it's confusing to me as well, I mean i couldn't recite all those rules concerning the authset/device lifecycle off the cuff...
anyway, good that we're having the discussion, I wouldn't have spotted it otherwise (or deemed it too important).

PR #213, opened 2020-09-30
MEN-3478: replace dot/dollar in inventory keys with unicode equivalents

tranchitella:
We need to replace dots and dollars in the key used to store inventory
attributes to allow mongodb successfully store and query them without
unnecessary nesting.

Changelog: none

### DIFF
--- a/model/device.go
+++ b/model/device.go
@@ -21,7 +21,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/go-ozzo/ozzo-validation/v4"
+	validation "github.com/go-ozzo/ozzo-validation/v4"
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/bson/bsontype"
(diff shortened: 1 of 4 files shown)


kacf (approved):
Nice!

I don't understand how the Unmarshalling works (you didn't change it, but I can see that it works), but that's just me not understanding the backend code, I think. :-)

model/device.go
```diff
@@ -39,6 +39,11 @@ const (
 	AttrNameCreated = "created_ts"
 )
 
+const (
+	runeDollar = '\uFF04'
+	runeDot    = '\uFF0E'
```
alfrunes, line 44: I was going to ask if this is right, but I see they're just the full width variants ． ＄.

alfrunes (approved):
LGTM

alfrunes:
Question: Will we need a schema migration here on existing documents? For existing instances there will pop up "duplicates" in the UI, but I don't think the client actually submit any attributes with dots and dollars in the default scripts.

model/device.go
tranchitella: yep

tranchitella:
@alfrunes in theory yes, we would need migration, but as we replaced dots with underscores, it is quite hard to reverse it as of today...

PR #214, opened 2020-10-02
MEN-4001: new device API end-point to replace inventory attributes

tranchitella:
Changelog: title


kjaskiewiczz (approved):
looks good

PR #219, opened 2020-10-09
Acceptance tests fail with INTERNALERROR 

merlin-northern:
ChangeLog:none


mzedel (approved):
that seems to be in line with some of the other repos... so: 🚂!!!

PR #223, opened 2020-10-23
MEN-3510: new API end-point to return the list of filterable attributes

tranchitella:
Changelog: title

### DIFF
--- a/api/http/api_inventory.go
+++ b/api/http/api_inventory.go
@@ -52,6 +52,7 @@ const (
 	urlInternalDevicesStatus = "/api/internal/v1/inventory/tenants/:tenant_id/devices/:status"
 	urlInternalAttributes    = "/api/internal/v1/inventory/tenants/:tenant_id/device/:did/attribute/scope/:scope"
 	apiUrlManagementV2       = "/api/management/v2/inventory"
+	urlFiltersAttributes     = apiUrlManagementV2 + "/filters/attributes"
 	urlFiltersSearch         = apiUrlManagementV2 + "/filters/search"
 
 	apiUrlInternalV2         = "/api/internal/v2/inventory"
@@ -118,6 +119,7 @@ func (i *inventoryHandlers) GetApp() (rest.App, error) {
 		rest.Post(uriInternalTenants, i.CreateTenantHandler),
 		rest.Post(uriInternalDevices, i.AddDeviceHandler),
 		rest.Post(urlInternalDevicesStatus, i.InternalDevicesStatusHandler),
+		rest.Get(urlFiltersAttributes, i.FiltersAttributesHandler),
 		rest.Post(urlFiltersSearch, i.FiltersSearchHandler),
 
 		rest.Post(urlInternalFiltersSearch, i.InternalFiltersSearchHandler),
@@ -743,6 +745,21 @@ func (i *inventoryHandlers) CreateTenantHandler(w rest.ResponseWriter, r *rest.R
 	w.WriteHeader(http.StatusCreated)
 }
 
+func (i *inventoryHandlers) FiltersAttributesHandler(w rest.ResponseWriter, r *rest.Request) {
+	ctx := r.Context()
+
+	l := log.FromContext(ctx)
+
+	// query the database
+	attributes, err := i.inventory.GetFiltersAttributes(ctx)
+	if err != nil {
+		u.RestErrWithLogInternal(w, r, l, err)
+		return
+	}
+
+	w.WriteJson(attributes)
+}
+
 func (i *inventoryHandlers) FiltersSearchHandler(w rest.ResponseWriter, r *rest.Request) {
 	ctx := r.Context()
 
@@ -755,7 +772,7 @@ func (i *inventoryHandlers) FiltersSearchHandler(w rest.ResponseWriter, r *rest.
 		return
 	}
 
-	// query database
+	// query the database
 	devs, totalCount, err := i.inventory.SearchDevices(ctx, *searchParams)
 	if err != nil {
 		if strings.Contains(err.Error(), "BadValue") {
@@ -788,7 +805,7 @@ func (i *inventoryHandlers) InternalFiltersSearchHandler(w rest.ResponseWriter, 
 		return
 	}
 
-	// query database
+	// query the database
 	devs, totalCount, err := i.inventory.SearchDevices(ctx, *searchParams)
 	if err != nil {
 		if strings.Contains(err.Error(), "BadValue") {
--- a/docs/management_api_v2.yml
+++ b/docs/management_api_v2.yml
@@ -32,6 +32,33 @@ securityDefinitions:
       Format: 'Bearer [JWT]'
 
 paths:
+  /filters/attributes:
+    get:
+      operationId: Get filterable attributes
+      tags:
+        - Management API
+      security:
+        - ManagementJWT: []
+      summary: Get the list of filterable inventory attributes
+      description:  |
+        Returns a list of filterable inventory attributes.
+
+        The list is sorted in descending order by the count of occurrences of the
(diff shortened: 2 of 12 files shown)


mzedel (approved):
👍 thank you very much for tackling this so quickly! I'll gladly follow up with the related changes in the gui repo! and yes that should work well for the gui to consume

api/http/api_inventory.go
```diff
@@ -743,6 +745,21 @@ func (i *inventoryHandlers) CreateTenantHandler(w rest.ResponseWriter, r *rest.R
 	w.WriteHeader(http.StatusCreated)
 }
 
+func (i *inventoryHandlers) FiltersAttributesHandler(w rest.ResponseWriter, r *rest.Request) {
+	ctx := r.Context()
+
+	l := log.FromContext(ctx)
+
+	// query database
```
merlin-northern, line 753: really nitpicking: 'query _the_ database'

docs/management_api_v2.yml
```diff
@@ -32,6 +32,33 @@ securityDefinitions:
...
+      tags:
+        - Management API
+      security:
+        - ManagementJWT: []
+      summary: Get the list of filterable inventory attributes
+      description:  |
+        Returns a list of filterable inventory attributes.
+
+        The list is sorted in descending order by the count of occurrences of the
+        attribute in the inventory database, then byin ascending order by scopeand name.
```
merlin-northern, line 47: nitpick redux: 'byin' -> 'by in'

merlin-northern: 'scopeand' -> 'scope and'

store/mongo/datastore_mongo.go
```diff
@@ -535,6 +535,69 @@ func (db *DataStoreMongo) UpdateDevicesGroup(
...
+	err = cur.All(ctx, &results)
+	if err != nil {
+		return nil, err
+	}
+
+	var attributes []model.FilterAttribute
+	for _, item := range results {
+		attributes = append(attributes, model.FilterAttribute{
+			Name:  item["_id"].(bson.M)[DbDevAttributesName].(string),
+			Scope: item["_id"].(bson.M)[DbDevAttributesScope].(string),
```
merlin-northern, line 593: we do have a const [DbDevId              = "_id"](https://github.com/mendersoftware/inventory/blob/master/store/mongo/datastore_mongo.go#L45)

merlin-northern (approved):
minor and just mean comments. looks good to me!

merlin-northern:
say what? `(-1.4%) to 43.408%` no way, you have to increase ;-)

tranchitella:
> say what? `(-1.4%) to 43.408%` no way, you have to increase ;-)

That's the acceptance, the overall coverage increased (+0.3%) to 86.897%

store/mongo/datastore_mongo.go
```diff
@@ -535,6 +535,71 @@ func (db *DataStoreMongo) UpdateDevicesGroup(
...
+		{
+			"$group": bson.M{
+				DbDevId: bson.M{
+					DbDevAttributesName:  "$" + DbDevAttributesName,
+					DbDevAttributesScope: "$" + DbDevAttributesScope,
+				},
+				DbCount: bson.M{
+					"$sum": 1,
+				},
+			},
```
alfrunes, line 570:
```suggestion
			"$group": bson.D{
				{Key: "_id", Value: nil},
				{Key: DbDevAttributesName, Value:  "$" + DbDevAttributesName},
				{Key: DbDevAttributesScope, Value: "$" + DbDevAttributesScope},
				DbCount: bson.M{
					"$sum": 1,
				},
			},
```
This way you don't have to do so many allocations when decoding the result.

merlin-northern:
lgtm

store/mongo/datastore_mongo.go
```diff
@@ -535,6 +535,71 @@ func (db *DataStoreMongo) UpdateDevicesGroup(
...
+		},
+		{
+			"$sort": bson.D{
```
alfrunes, line 573: I was going to comment on the `bson.M` here before you fixed it, but generally I think it's a good habit to discourage use of `bson.M` everywhere as it can have unintended random side-effects elsewhere as well.

store/mongo/datastore_mongo.go
```diff
@@ -535,6 +535,71 @@ func (db *DataStoreMongo) UpdateDevicesGroup(
...
+
+	var attributes []model.FilterAttribute
+	for _, item := range results {
+		attributes = append(attributes, model.FilterAttribute{
+			Name:  item[DbDevId].(bson.M)[DbDevAttributesName].(string),
+			Scope: item[DbDevId].(bson.M)[DbDevAttributesScope].(string),
+			Count: item[DbCount].(int32),
+		})
+	}
+
```
alfrunes, line 599: This is not ideal.
If you reformat the `$group` stage in the pipeline you can decode this directly from the cursor.

store/mongo/datastore_mongo.go
```diff
@@ -535,6 +535,71 @@ func (db *DataStoreMongo) UpdateDevicesGroup(
...
+		},
+		{
+			"$unwind": "$" + DbDevAttributes,
+		},
+		{
+			"$project": bson.M{
+				DbDevAttributesName:  "$" + DbDevAttributes + ".v." + DbDevAttributesName,
+				DbDevAttributesScope: "$" + DbDevAttributes + ".v." + DbDevAttributesScope,
+			},
+		},
```
alfrunes, line 560: Should we add a `$skip` and limit stage `$limit` here?
The `$group` stage is allowed to allocate up to 100MiB of RAM before it returns an error, it might be safer to add pagination to the API and limit the usage?

alfrunes (approved):
PR looks fine.

I only have a couple loose questions and suggestions.

store/mongo/datastore_mongo.go
tranchitella: @alfrunes thanks for the suggestion, you are right

store/mongo/datastore_mongo.go
tranchitella: @alfrunes good question, I believe pagination is not going to be very efficient, though, as we have to run the aggregation over and over. I'm more in favor of a fixed $limit (e.g. `500`), as the main usage of this end-point is to provide autocompletion in the UI. What do you think?

store/mongo/datastore_mongo.go
tranchitella: Well, this does not work: in mongodb, only the `_id` field is used for grouping fields, all other fields need to specify an accumulator (https://docs.mongodb.com/manual/reference/operator/aggregation/group/). Applying your change, I get:

The field 'name' must be an accumulator object

store/mongo/datastore_mongo.go
tranchitella: Well noted

store/mongo/datastore_mongo.go
tranchitella: I've done it using an extra step in the pipeline using $project, which allows us to deserialize the cursor directly to the array of model.FilterAttribute.

alfrunes: Ahh, ok. I see I misunderstood the docs for `$group`.
I guess another option without adding an extra step to the pipeline would be creating an alias
```
type MongoFilterAttr model.FilterAttribute

func (f *MongoFilterAttr) UnmarshalBSON(b []byte) error {
       ... 
} 
```

store/mongo/datastore_mongo.go
alfrunes: Hmm.. I guess a fixed limit makes more sense when you put it like that. It would be pretty awkward if the endpoint suddenly stopped working above a certain device limit, I guess we can come back and tweak the exact limit at a later stage if we feel like it but 500 sounds plenty.

PR #232, opened 2020-12-20
Index commonly used inventory attributes' values

tranchitella:
Changelog: none

### DIFF
--- a/store/mongo/datastore_mongo.go
+++ b/store/mongo/datastore_mongo.go
@@ -37,7 +37,7 @@ import (
 )
 
 const (
-	DbVersion = "1.0.0"
+	DbVersion = "1.0.1"
 
 	DbName        = "inventory"
 	DbDevicesColl = "devices"
@@ -58,6 +58,8 @@ const (
 	DbScopeInventory = "inventory"
 
 	FiltersAttributesLimit = 500
+
+	attrIdentityStatus = "identity-status"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


tranchitella:
QA pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/232394904

store/mongo/datastore_mongo.go
```diff
@@ -989,10 +991,15 @@ func (db *DataStoreMongo) SearchDevices(ctx context.Context, searchParams model.
 func indexAttr(s *mongo.Client, ctx context.Context, attr string) error {
 	l := log.FromContext(ctx)
 	c := s.Database(mstore.DbFromContext(ctx, DbName)).Collection(DbDevicesColl)
-	indexField := fmt.Sprintf("attributes.%s.values", attr)
 
 	indexView := c.Indexes()
-	_, err := indexView.CreateOne(ctx, mongo.IndexModel{Keys: bson.M{indexField: 1}, Options: nil})
+	keys := bson.D{
+		{Key: indexAttrName(attrIdentityStatus), Value: 1},
```
kjaskiewiczz, line 997: are we using different attributes always in combination with "identity-status" in queries?

tranchitella: Yes, all queries from UI are "status = something" in combination with the other criterias:
* latest update time
* group name
* mac address
* device type
We won't cover API-based queries that do not filter by status, but as it makes sense to act only on active devices when dynamically filtering, I believe we can assume it will always be there.

kjaskiewiczz (approved):
looks good

PR #275, opened 2021-06-23
internal endpoint for getting the device's groups

kjaskiewiczz:


kjaskiewiczz:

api/http/api_inventory.go
```diff
@@ -50,6 +50,7 @@ const (
 	uriInternalTenants       = "/api/internal/v1/inventory/tenants"
 	uriInternalDevices       = "/api/internal/v1/inventory/devices"
 	urlInternalDevicesStatus = "/api/internal/v1/inventory/tenants/:tenant_id/devices/:status"
+	uriInternalDeviceGroup   = "/api/internal/v1/inventory/tenants/:tenant_id/devices/:status/group"
```
tranchitella, line 53: Why `:status`? Cannot it be `:device_id`?

kjaskiewiczz: it cannot be device_id because we are already using "/api/internal/v1/inventory/tenants/:tenant_id/devices/:status"
if we'll use device_id it will cause runtime error:
 _inventory API handlers setup failed: failed to create router: Routes sharing a common placeholder MUST name it consistently: status != device_id_

kjaskiewiczz: I can change /api/internal/v1/inventory/tenants/:tenant_id/devices/:status to /api/internal/v1/inventory/tenants/:tenant_id/devices/status/:status or /api/internal/v1/inventory/tenants/:tenant_id/devices/status and move status to the body
update workflows or workflows and deviceauth
and then use device_id here
/api/internal/v1/inventory/tenants/:tenant_id/devices/:device_id/group

tranchitella: Oh, I didn't know the framework wants to have common prefixes.
I don't think moving the internal endpoint is a problem, but it is additional work.
Feel free to create a ticket for it, we can follow-up later without blocking your work in this PR.

api/http/api_inventory.go
```diff
@@ -899,6 +901,42 @@ func (i *inventoryHandlers) InternalDevicesStatusHandler(w rest.ResponseWriter,
 	w.WriteJson(result)
 }
 
+func (i *inventoryHandlers) GetDeviceGroupInternalHandler(w rest.ResponseWriter, r *rest.Request) {
+	ctx := r.Context()
+
+	l := log.FromContext(ctx)
+
+	tenantId := r.PathParam("tenant_id")
+	if tenantId == "" {
```
tranchitella, line 910: I know this end-point is needed by enterprise packages, but is it really needed to make this end-point multi-tenant only? In other places in the code, we accept an empty tenantID.

tranchitella (approved):

api/http/api_inventory.go
kjaskiewiczz: fixed; this was mistake done when backporting from enterprise

kjaskiewiczz:

kjaskiewiczz:
@tranchitella I have to ask for one more review here;
the main difference since your last review is that the new endpoint returns **list** of groups (it's a list with a single group for now, but it may return more groups in the future) instead of single group

tranchitella (approved):
Good idea to return an array!

PR #278, opened 2021-07-02
MEN-4691: management endpoint for updating device tags

0lmi:
https://tracker.mender.io/browse/MEN-4691
https://tracker.mender.io/browse/MEN-4692


tests/tests/test_mgmt_tag_attributes.py
```diff
@@ -0,0 +1,165 @@
...
+        did = "some-device-id"
+        tags = {"n_5": {"name": "n_5", "value": "v_5", "description": "desc_5"}}
+        tags_body = [tags['n_5']]
+        internal_client.create_device(did, inventory_attributes)
+        management_client.updateTagAttributes(did, tags_body)
+
+        fake_etag = "241496e0-cbbb-4a83-90e9-70b4dd0e645a"
+        try:
+            management_client.updateTagAttributes(did, tags_body, eTag=fake_etag)
+        except Exception as e:
```
kjaskiewiczz, line 147: what if there will be no exception?

store/mongo/datastore_mongo.go
```diff
@@ -368,22 +375,47 @@ func (db *DataStoreMongo) upsertAttributes(
 	case 1:
 		var res *mongo.UpdateResult
 		if withRevision {
-			filter = bson.M{
-				"_id":         devices[0].Id,
-				DbDevRevision: bson.M{"$lt": devices[0].Revision},
+			if etag != nil && *etag != "" {
+				filter = bson.M{
```
kjaskiewiczz, line 379: I think it should be possible to build filter in a different way to avoid nested checks
e.g. https://github.com/mendersoftware/deployments/blob/master/store/mongo/datastore_mongo.go#L1794

store/mongo/datastore_mongo.go
```diff
@@ -368,22 +375,47 @@ func (db *DataStoreMongo) upsertAttributes(
...
-			filter = map[string]interface{}{"_id": devices[0].Id}
-			update = bson.M{
-				"$set":         update,
-				"$setOnInsert": oninsert,
+			if etag != nil && *etag != "" {
+				filter = bson.M{
+					"_id":     devices[0].Id,
+					etagField: bson.M{"$exists": true, "$eq": *etag},
+				}
+				// check if we are going to update existing document with the
```
kjaskiewiczz, line 401: I think, even with this check, it will still be possible to get duplicate key error under some conditions (with concurrent updates) and it looks like we have a handling for this kind of errors, but for a different case, which makes it hard to distinguish both cases (maybe using a scope to distinguish them would be an option?)

kjaskiewiczz (commented):
I left some comments, sorry if they are messy (I didn't have enough time to spend with this PR, I will try to take one more look asap)

tests/tests/test_mgmt_tag_attributes.py
```diff
@@ -0,0 +1,165 @@
...
+        tags = {"n_6": {"name": "n_6", "value": "v_6", "description": "desc_6"}}
+        tags_body = [tags['n_6']]
+        internal_client.create_device(did, inventory_attributes)
+        management_client.setTagAttributes(did, tags_body)
+
+        fake_etag = "241496e0-cbbb-4a83-90e9-70b4dd0e645a"
+        try:
+            management_client.setTagAttributes(did, tags_body, eTag=fake_etag)
+        except Exception as e:
+            assert str(e) == "412 Precondition Failed"
```
tranchitella, line 165:
```suggestion
        try:
            management_client.setTagAttributes(did, tags_body, eTag=fake_etag)
        except Exception as e:
            assert str(e) == "412 Precondition Failed"
        else:
            raise Exception("did not raise expected exception")
```

api/http/api_inventory.go
```diff
@@ -384,37 +388,85 @@ func (i *inventoryHandlers) AddDeviceHandler(w rest.ResponseWriter, r *rest.Requ
...
+}
+
+func (i *inventoryHandlers) updateDeviceAttributes(
+	w rest.ResponseWriter,
+	r *rest.Request,
+	ctx context.Context,
+	attrs model.DeviceAttributes,
+	deviceID model.DeviceID,
+	scope string,
+	etag *string,
```
tranchitella, line 447: Why is etag a pointer? I believe a string is enough, and you can check if not empty instead of `!= nil`

docs/management_api.yml
```diff
@@ -198,6 +206,121 @@ paths:
...
+      summary: Adds a set of tags for a device
+      description: |
+        Saves the provided tags set for the authenticated device.
+        The device ID is retrieved from the authorization header.
+
+        This method has upsert semantics:
+
+        * the values of existing attributes are overwritten
+
+        * attributes assigned for the first time are automatically created
```
tranchitella, line 225:
```suggestion
        This method has upsert semantic:
        * it overwrites the values of existing attributes
        * it creates attributes assigned for the first time
```

docs/management_api.yml
```diff
@@ -198,6 +206,121 @@ paths:
...
+      parameters:
+        - name: If-Match
+          in: header
+          description: Contains ETag of modified device object.
```
tranchitella, line 229:
```suggestion
          description: Contains the device object's current ETag, and performs the update only if it matches the one stored in the database.
```

docs/management_api.yml
```diff
@@ -198,6 +206,121 @@ paths:
...
+      description: |
+        Replaces tags to the provided set for the authenticated device.
+
+        This method replaces all the tags with the new set:
+
+        * attributes not provided in the set are removed from the db
+
+        * the values of existing attributes are overwritten
+
+        * attributes assigned for the first time are automatically created
```
tranchitella, line 282:
```suggestion
        Replaces the tags associated to the device.

        This method replaces all the tags with the new set:

        * it removes from the db the attributes not provided
        * it overwrites the values of existing attributes
        * it creates attributes assigned for the first time
```

docs/management_api.yml
```diff
@@ -198,6 +206,121 @@ paths:
...
+      parameters:
+        - name: If-Match
+          in: header
+          description: Contains ETag of modified device object.
```
tranchitella, line 286:
```suggestion
          description: Contains the device object's current ETag, and performs the update only if it matches the one stored in the database.
```

store/mongo/datastore_mongo.go
```diff
@@ -368,22 +375,47 @@ func (db *DataStoreMongo) upsertAttributes(
...
-			}
 		} else {
-			filter = map[string]interface{}{"_id": devices[0].Id}
-			update = bson.M{
-				"$set":         update,
-				"$setOnInsert": oninsert,
+			if etag != nil && *etag != "" {
+				filter = bson.M{
+					"_id":     devices[0].Id,
+					etagField: bson.M{"$exists": true, "$eq": *etag},
```
tranchitella, line 399: $eq is enough, no need to use $exists as well

store/mongo/datastore_mongo.go
```diff
@@ -368,22 +375,47 @@ func (db *DataStoreMongo) upsertAttributes(
...
+				// with 'duplicate key' error
+				if err := checkMatchedCount(ctx, c, filter); err != nil {
+					return &model.UpdateResult{MatchedCount: 0}, nil
+				}
+			} else {
+				filter = map[string]interface{}{"_id": devices[0].Id}
 			}
 		}
+		if scope == model.AttrScopeTags {
+			update[etagField] = fmt.Sprintf("%s", uuid.New())
```
tranchitella, line 412: uuid objects have a method to return a string (is it .String(), maybe?)

inv/inventory.go
```diff
@@ -161,9 +178,16 @@ func (i *inventory) ReplaceAttributes(ctx context.Context, id model.DeviceID, up
 			}
 		}
 	}
-	if _, err := i.db.UpsertRemoveDeviceAttributes(ctx, id, upsertAttrs, removeAttrs); err != nil {
+
+	res, err := i.db.UpsertRemoveDeviceAttributes(ctx, id, upsertAttrs, removeAttrs, scope, etag)
```
tranchitella, line 182: Pay attention: this store method also sets updated_ts, which we do not want to set when changing the tags.
The reason is that the updated_ts is also known as "latest check-in", and we use that as a cheap monitor solution for the devices. If you set the value when changing tags, we lose the information.

store/mongo/datastore_mongo.go
```diff
@@ -558,6 +612,26 @@ func (db *DataStoreMongo) UpsertRemoveDeviceAttributes(
 		Database(mstore.DbFromContext(ctx, DbName)).
 		Collection(DbDevicesColl)
 
+	var filter interface{}
+	if etag != nil && *etag != "" {
+		filter = bson.M{
+			"_id":     id,
+			etagField: bson.M{"$exists": true, "$eq": *etag},
```
tranchitella, line 619: Again, no need to use $exists and $eq together

api/http/api_inventory.go
```diff
@@ -384,37 +388,85 @@ func (i *inventoryHandlers) AddDeviceHandler(w rest.ResponseWriter, r *rest.Requ
...
 		return
 	}
+	i.updateDeviceAttributes(w, r, ctx, attrs, deviceID, model.AttrScopeInventory, nil)
+}
+
+func (i *inventoryHandlers) UpdateDeviceTagsHandler(w rest.ResponseWriter, r *rest.Request) {
+	ctx := r.Context()
+	l := log.FromContext(ctx)
+
+	// get device ID from uri depending from a scope
```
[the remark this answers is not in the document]
0lmi: Yes, the comment was written before refactoring and now irrelevant. Thanks!

tests/tests/test_mgmt_tag_attributes.py
0lmi: @tranchitella I thought it is enough to fail on assert in this case. It looks something like:
```
tests/test_mgmt_tag_attributes.py:165: in test_replace_tag_attributes_with_wrong_etag
     assert str(e) == "412 Precondition Failed"
 E   AssertionError: assert '500 Internal Server Error' == '412 Precondition Failed'
 E     - 412 Precondition Failed
 E     + 500 Internal Server Error
```

inv/inventory.go
0lmi: Good point. Thanks!

store/mongo/datastore_mongo.go
0lmi: Makes sense. Thanks!

store/mongo/datastore_mongo.go
0lmi: Great idea! Thanks!

store/mongo/datastore_mongo.go
0lmi: Yes. Good point. Thanks!

tests/tests/test_mgmt_tag_attributes.py
0lmi: Ah, it's for the case when no any exception raised. Makes sense. Thanks!

store/mongo/datastore_mongo.go
0lmi: Makes sense

store/mongo/datastore_mongo.go
0lmi: 👍

api/http/api_inventory.go
0lmi: Makes sense. Looks like left over from the beginning and not doesn't seems like it's needed. Thanks.

store/mongo/datastore_mongo.go
0lmi: Initially I thought that building logic based on errors is not the best, but in this case it looks way better than extra DB call + not handled cases.

inv/inventory.go
```diff
@@ -161,7 +181,12 @@ func (i *inventory) ReplaceAttributes(ctx context.Context, id model.DeviceID, up
 			}
 		}
 	}
-	if _, err := i.db.UpsertRemoveDeviceAttributes(ctx, id, upsertAttrs, removeAttrs); err != nil {
+
+	_, err = i.db.UpsertRemoveDeviceAttributes(ctx, id, upsertAttrs, removeAttrs, scope, etag)
+	if err != nil {
+		if scope == model.AttrScopeTags {
+			return ErrETagDoesntMatch
```
kjaskiewiczz, line 188: return this error only on write conflicts, not on any error

inv/inventory.go
```diff
@@ -131,16 +135,31 @@ func (i *inventory) UpsertAttributes(ctx context.Context, id model.DeviceID, att
...
+	attrs model.DeviceAttributes,
+	scope string,
+	etag string,
+) error {
+	_, err := i.db.UpsertDevicesAttributesWithUpdated(
+		ctx, []model.DeviceID{id}, attrs, scope, etag,
+	)
+	if err != nil {
+		if scope == model.AttrScopeTags {
+			return ErrETagDoesntMatch
```
kjaskiewiczz, line 150: return this error only on write conflicts, not on any error

inv/inventory.go
```diff
@@ -131,16 +135,31 @@ func (i *inventory) UpsertAttributes(ctx context.Context, id model.DeviceID, att
...
+		}
 		return errors.Wrap(err, "failed to upsert attributes in db")
 	}
+	// if etag != nil && *etag != "" {
```
kjaskiewiczz, line 154: commented code

inv/inventory.go
```diff
@@ -131,16 +135,31 @@ func (i *inventory) UpsertAttributes(ctx context.Context, id model.DeviceID, att
...
+	// 	if res != nil && res.MatchedCount == 0 {
+	// 		return ErrETagDoesntMatch
+	// 	}
+	// }
```
tranchitella, line 158: Why do we need this commented code?

model/device.go
```diff
@@ -144,6 +148,9 @@ type Device struct {
 
 	//device object revision
 	Revision uint `json:"-" bson:"revision,omitempty"`
+
+	//tags attributes ETag
+	Tags_etag string `json:"tags_etag,omitempty" bson:"tags_etag,omitempty"`
```
tranchitella, line 153:
```suggestion
	TagsETag string `json:"tags_etag,omitempty" bson:"tags_etag,omitempty"`
```

store/mongo/datastore_mongo.go
```diff
@@ -567,11 +591,20 @@ func (db *DataStoreMongo) UpsertRemoveDeviceAttributes(
...
-		Value: now,
+	filter := bson.M{"_id": id}
+	if etag != "" {
+		filter[etagField] = bson.M{"$eq": etag}
+	} else {
+		update[updatedField] = model.DeviceAttribute{
+			Scope: model.AttrScopeSystem,
+			Name:  model.AttrNameUpdated,
+			Value: now,
+		}
```
tranchitella, line 607: This is wrong, I believe: the ETag is not mandatory, you cannot rely on it to ste the "updated at" field.
If I set tags but I do not provide the If-Match, you are still updating the timestamp.

tranchitella (changes requested):

inv/inventory.go
0lmi: this is mistake, removing

store/mongo/datastore_mongo.go
0lmi: You are right, counting on etag here is wrong, scope sounds the right criteria.

inv/inventory.go
```diff
@@ -161,7 +179,14 @@ func (i *inventory) ReplaceAttributes(ctx context.Context, id model.DeviceID, up
 			}
 		}
 	}
-	if _, err := i.db.UpsertRemoveDeviceAttributes(ctx, id, upsertAttrs, removeAttrs); err != nil {
+
+	_, err = i.db.UpsertRemoveDeviceAttributes(ctx, id, upsertAttrs, removeAttrs, scope, etag)
+	if err != nil {
+		if scope == model.AttrScopeTags {
+			if strings.Contains(err.Error(), "E11000 duplicate key error collection") {
```
0lmi, line 186: @tranchitella @kjaskiewiczz I'm unsure about good enough productivity in this approach. Any ideas to make it more efficient?

0lmi:

kjaskiewiczz (approved):
looks good to me

tranchitella (approved):

PR #291, opened 2021-08-31
MEN-4525: improve device decommissioning and inventory data reset

tranchitella:
Introduce a new internal end-point to delete (decommission) a device
from the inventory microservice. The new end-point replaces an
equivalent one in the management APIs we deprecated because there was no
reason for users to manually trigger the decommissioning for a device.
At the same time, update the implementation of the old management
endpoint (DELETE) with a new semantic: reset the device's inventory
data, removing all the attributes in the inventory scope. This is what
the end-point was reallly about in its original intentions.

Changelog: title


tranchitella:
Combined QA: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/362507133

alfrunes (approved):

PR #297, opened 2021-09-30
MEN-4644: add support for filtering datetimes (e.g. `updated_ts`)

tranchitella:
Changelog: title


kjaskiewiczz (approved):
looks good

PR #300, opened 2021-10-15
Use the `hdrTotalCount` constant everywhere

tranchitella:
Changelog: none


merlin-northern (approved):
lgtm. as a review I grepped for all occurences and I confirm these are all.

alfrunes (approved):

PR #301, opened 2021-10-17
MEN-4826: new API end-point to remove a static group

tranchitella:


tranchitella:

store/mongo/datastore_mongo.go
```diff
@@ -740,6 +740,45 @@ func (db *DataStoreMongo) GetFiltersAttributes(ctx context.Context) ([]model.Fil
...
+				res, err := collDevs.UpdateOne(ctx, bson.M{DbDevId: device.ID}, update)
+				if err == nil && res.MatchedCount == 1 {
+					updateDeviceID = device.ID
+				}
+				deviceIDs <- updateDeviceID
+			}
+		}
+
+		close(deviceIDs)
+	}()
```
alfrunes, line 777: Yikes! This is not very efficient. We should also use batch operations (`UpdateMany`) for mongo as well, since we expect all the device IDs to match a single document (even though it's not necessarily always the case).

store/mongo/datastore_mongo.go
```diff
@@ -740,6 +740,45 @@ func (db *DataStoreMongo) GetFiltersAttributes(ctx context.Context) ([]model.Fil
...
+	cursor, err := collDevs.Find(ctx, filter, findOptions)
+	if err != nil {
+		return nil, err
+	}
+	go func() {
+		defer cursor.Close(ctx)
+
+		update := bson.M{"$unset": bson.M{DbDevAttributesGroup: 1}}
+		for cursor.Next(ctx) {
+			device := &model.Device{}
```
alfrunes, line 765: Unnecessary memory allocation inside for-loop.

inv/inventory.go
```diff
@@ -323,6 +324,28 @@ func (i *inventory) GetFiltersAttributes(ctx context.Context) ([]model.FilterAtt
...
+
+	res := &model.UpdateResult{}
+	for deviceID := range deviceIDs {
+		deviceIDStr := string(deviceID)
+		res.MatchedCount += 1
+		if deviceIDStr != "" {
+			res.UpdatedCount += 1
+			i.maybeTriggerReindex(ctx, deviceIDStr)
+		}
+	}
```
alfrunes, line 344: Can we check if reporting exist outside instead of inside the loop?
```suggestion
	if i.enableReporting {
		res := &model.UpdateResult{}
		for deviceID := range deviceIDs {
			deviceIDStr := string(deviceID)
			res.MatchedCount += 1
			if deviceIDStr != "" {
				res.UpdatedCount += 1
				i.triggerReindex(ctx, deviceIDStr)
			}
		}
	}
```

store/mongo/datastore_mongo.go
```diff
@@ -740,6 +740,45 @@ func (db *DataStoreMongo) GetFiltersAttributes(ctx context.Context) ([]model.Fil
...
+
+	filter := bson.M{DbDevAttributesGroupValue: group}
+	findOptions := &mopts.FindOptions{
+		Projection: bson.M{DbDevId: 1},
+	}
+	cursor, err := collDevs.Find(ctx, filter, findOptions)
+	if err != nil {
+		return nil, err
+	}
+	go func() {
```
alfrunes, line 760: Should we check if reporting exist before forking this go routine?

store/mongo/datastore_mongo.go
tranchitella: My problem with `UpdateMany` is that we won't get back the device IDs, which means that we cannot reindex them in ES (reporting service).

store/mongo/datastore_mongo.go
tranchitella: I don't think so, and the datastore doesn't know anything about reporting.

PR #323, opened 2022-03-29
fix: device search reindexing

tranchitella:
Changelog: none
Ticket: MEN-5527

### DIFF
--- a/store/mongo/datastore_mongo.go
+++ b/store/mongo/datastore_mongo.go
@@ -606,8 +606,7 @@ func (db *DataStoreMongo) UpsertRemoveDeviceAttributes(
 	const createdField = systemScope + "-" + model.AttrNameCreated
 	const etagField = model.AttrNameTagsEtag
 	var (
-		result *model.UpdateResult
-		err    error
+		err error
 	)
 
 	c := db.client.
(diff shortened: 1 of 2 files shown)


tranchitella:
I triggered the QA pipeline manually: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/503613504

tranchitella:
We can ignore mender-connect failures as they are unrelated.

store/mongo/datastore_mongo.go
```diff
@@ -654,15 +655,23 @@ func (db *DataStoreMongo) UpsertRemoveDeviceAttributes(
...
 
-	var res *mongo.UpdateResult
-	res, err = c.UpdateOne(ctx, filter, update, updateOpts)
-	if err == nil {
-		result = &model.UpdateResult{
-			MatchedCount: res.MatchedCount,
-			CreatedCount: res.UpsertedCount,
-		}
+	device := &model.Device{}
+	res := c.FindOneAndUpdate(ctx, filter, update, updateOpts)
```
alfrunes, line 659: I don't get why we need to return the device document. Isn't this just extra overhead?

alfrunes (approved):

store/mongo/datastore_mongo.go
tranchitella: We need to return the document to (eventually) reindex the full-text field.

It is technically possible to infer the full-text field without retrieving the document after the update, but this can lead to race conditions and we wanted to come up with a simple solution, at least at this stage. Therefore, we update inventory attributes/tags retrieving the document after the update, compose the full-text field and update it with an independent query if different from the one we retrieved from the db.

PR #357, opened 2022-09-06
fix(api-docs): type for the value attribute in the devices API docs

tpoxa:
As the mender client may provide a string or array of strings for an attribute value,  *object* is a more appropriate type for Swagger 2.0.

### DIFF
--- a/docs/devices_api.yml
+++ b/docs/devices_api.yml
@@ -139,7 +139,7 @@ definitions:
         type: string
         description: Attribute description.
       value:
-        type: string
+        type: object
         description: |
             The current value of the attribute.
 


tranchitella:
Thanks for your contribution @tpoxa. Can I kindly ask you to amend your commit to:
- Sign-off your commit (you can use the "-s" git option)
- Change the commit message to something like:

```
fix(api-docs): type for the value attribute in the devices API docs

As the mender client may provide a string or array of strings for an attribute value, *object* is a more appropriate type for Swagger 2.0.

Changelog: Title
Ticket: None
```

PR #369, opened 2022-12-19
chore: dependabot updates to the vendorized libraries

tranchitella:
Changelog: None
Ticket: None


merlin-northern (approved):
as a review I cloned, removed vendor, ran go mod tidy and go mod vendor. looks good to me.

PR #380, opened 2023-03-25
Dependabot updates

tranchitella:


Dockerfile.acceptance-testing
```diff
@@ -14,6 +15,7 @@ RUN env CGO_ENABLED=0 go test -c -o inventory -tags main \
 FROM scratch
 EXPOSE 8080
 COPY --from=builder /etc_extra/ /etc/
+COPY --from=builder --chown=nobody /tmp_extra/ /tmp/
```
merlin-northern, line 18: why are we coping tmp_extra to tmp?

merlin-northern (changes requested):

Dockerfile.acceptance-testing
merlin-northern: ok I see it, sorry for the noise.

merlin-northern (approved):

PR #386, opened 2023-05-17
fix value returned by /filters/attributes endpoint

kjaskiewiczz:
Ticket: MEN-6477

In the /filters/attributes endpoint, in case there are no attributes,
return empty list instead of null.


kjaskiewiczz:
@mender-test-bot start pipeline --fast

merlin-northern (approved):
lgtm!

PR #387, opened 2023-05-18
automatically check if the response from /filters/attributes is aligned with API specification

kjaskiewiczz:


tests/tests/client.py
```diff
@@ -147,6 +147,33 @@ def addDeviceToGroup(self, group, device, expected_error=False, JWT="foo.bar.baz
...
+        }
+
+        http_client = RequestsClient()
+        http_client.session.verify = False
+
+        self.client = SwaggerClient.from_spec(
+            load_file(spec), config=config, http_client=http_client,
+        )
+        self.client.swagger_spec.api_url = (
+            "http://%s/api/management/v2/inventory/" % host
```
merlin-northern, line 169: nit pick dont we have it defined in some cost somewhere? (I have not checked)

merlin-northern (approved):
lgtm

tests/tests/client.py
kjaskiewiczz: with openapi it's completely different, but bravado is a bit clunky, and this is what we do in other places where we need v2 API

tests/tests/test_mgmt_v2_filters_attributes.py
```diff
@@ -0,0 +1,50 @@
...
+        )
+        attributeList.append(attr)
+
+        did = "some-device-id"
+        internal_client.create_device(did, attributeList)
+        res = management_client_v2.getFiltersAttributes()
+        for attr in res:
+            if attr.scope == "inventory":
+                assert res[0].name == "foo"
+                assert res[0].count == 1
```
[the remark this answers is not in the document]
mzedel: Thanks for the clarification offline @kjaskiewiczz !

PR #396, opened 2023-06-29
feat: update inventory only when changed or outdated.

merlin-northern:
Changelog: Title
Ticket: MEN-6425

### DIFF
--- a/model/device.go
+++ b/model/device.go
@@ -146,6 +146,48 @@ func allowedType(v reflect.Value) bool {
 	return false
 }
 
+func reflectValuesEqual(rVal1 reflect.Value, rVal2 reflect.Value) bool {
+	if rVal1.Len() != rVal2.Len() {
+		return false
+	}
+	for i := 0; i < rVal1.Len(); i++ {
+		if rVal1.Index(i).Kind() != rVal2.Index(i).Kind() {
+			return false
+		}
+		if !allowedType(rVal1.Index(i)) || !allowedType(rVal2.Index(i)) {
+			if !reflect.DeepEqual(rVal1.Index(i), rVal2.Index(i)) {
+				return false
+			}
+		} else {
+			if rVal2.Index(i).Kind() == reflect.Int {
+				value1 := rVal1.Index(i).Interface().(int)
+				value2 := rVal2.Index(i).Interface().(int)
+				if value1 != value2 {
+					return false
+				}
+			}
+			if rVal2.Index(i).Kind() == reflect.String {
+				value1 := rVal1.Index(i).Interface().(string)
+				value2 := rVal2.Index(i).Interface().(string)
+				if value1 != value2 {
+					return false
+				}
+			}
+			if rVal2.Index(i).Kind() == reflect.Float64 {
+				floatComparePrecision := "%.8f"
+				floatValue1 := rVal1.Index(i).Interface().(float64)
+				floatValue2 := rVal2.Index(i).Interface().(float64)
+				value1 := fmt.Sprintf(floatComparePrecision, floatValue1)
+				value2 := fmt.Sprintf(floatComparePrecision, floatValue2)
+				if value1 != value2 {
+					return false
+				}
+			}
+		}
+	}
+	return true
+}
+
 func (da DeviceAttribute) Equal(e DeviceAttribute) bool {
 	if (da.Value == nil || e.Value == nil) && !(da.Value == nil && e.Value == nil) {
 		return false
@@ -160,49 +202,7 @@ func (da DeviceAttribute) Equal(e DeviceAttribute) bool {
 		}
 	}
 	if rVal1.Kind() == reflect.Slice || rVal1.Kind() == reflect.Array {
-		if rVal1.Len() != rVal2.Len() {
-			return false
-		}
-		for i := 0; i < rVal1.Len(); i++ {
-			//if !rVal1.Comparable() || !rVal1.Index(i).Equal(rVal2.Index(i)) { //does not work:
-			//nolint:lll
-			//model/device.go:149:14: rVal1.Comparable undefined (type reflect.Value has no field or method Comparable)
-			//nolint:lll
-			//model/device.go:149:46: rVal1.Index(i).Equal undefined (type reflect.Value has no field or method Equal)
-			if rVal1.Index(i).Kind() != rVal2.Index(i).Kind() {
-				return false
-			}
-			if !allowedType(rVal1.Index(i)) || !allowedType(rVal2.Index(i)) {
-				if rVal1.Index(i) != rVal2.Index(i) {
-					return false
-				}
-			} else {
-				if rVal2.Index(i).Kind() == reflect.Int {
-					value1 := rVal1.Index(i).Interface().(int)
-					value2 := rVal2.Index(i).Interface().(int)
-					if value1 != value2 {
-						return false
-					}
-				}
-				if rVal2.Index(i).Kind() == reflect.String {
-					value1 := rVal1.Index(i).Interface().(string)
-					value2 := rVal2.Index(i).Interface().(string)
-					if value1 != value2 {
-						return false
-					}
-				}
-				if rVal2.Index(i).Kind() == reflect.Float64 {
-					floatComparePrecision := "%.8f"
-					floatValue1 := rVal1.Index(i).Interface().(float64)
-					floatValue2 := rVal2.Index(i).Interface().(float64)
-					value1 := fmt.Sprintf(floatComparePrecision, floatValue1)
-					value2 := fmt.Sprintf(floatComparePrecision, floatValue2)
-					if value1 != value2 {
-						return false
-					}
-				}
-			}
-		}
+		return reflectValuesEqual(rVal1, rVal2)
 	} else {
 		if da.Value != e.Value {
 			return false
@@ -224,7 +224,7 @@ func (da DeviceAttribute) Equal(e DeviceAttribute) bool {
 			}
 		}
 	}
-	return true
+	return false
 }
 
 func validateDeviceAttrVal(i interface{}) error {
--- a/model/device.go
+++ b/model/device.go
@@ -1,4 +1,4 @@
-// Copyright 2022 Northern.tech AS
+// Copyright 2023 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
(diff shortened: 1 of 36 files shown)


merlin-northern:
@mender-test-bot start pipeline --fast please

merlin-northern:
@alfrunes I need a review while I add tests

model/device.go
```diff
@@ -69,6 +69,64 @@ type DeviceAttribute struct {
...
+			}
+		}
+	}
+
+	// cannot be equal if keys differ (the last possibility)
+	for k := range attributes {
+		if _, ok := existingAttributes[k]; !ok {
+			return false
+		}
+	}
```
alfrunes, line 125: You can collapse the second loop in the first one.
```suggestion
	for k, v := range attributes {
		if e, ok := existingAttributes[k]; ok {
			if !v.Equal(e) {
				return false
			}
		} else {
			return false
		}
	}
```

model/device.go
```diff
@@ -78,6 +136,37 @@ func (da DeviceAttribute) Validate() error {
...
+	ok1 := reflect.TypeOf(da.Value).Kind() == reflect.Slice
+	ok2 := reflect.TypeOf(e.Value).Kind() == reflect.Slice
+	if ok1 || ok2 {
+		// there is no portable way to compare slices of unknown type
+		return false
+	} else {
+		if da.Value != e.Value {
+			return false
+		}
+	}
```
alfrunes, line 152: > // there is no portable way to compare slices of unknown type

Yes, you can...
```suggestion
	rVal1 := reflect.ValueOf(v1)
	rVal2 := reflect.ValueOf(v2)
	if rVal1.Kind() != rVal2.Kind() {
		return false
	}
	if rVal1.Kind() == reflect.Slice {
		if rVal1.Len() != rVal2.Len() {
			return false
		}
		elemKind := rVal1.Elem().Kind()
		if elemKind != rVal2.Elem().Kind() {
			return false
		} else if elemKind != reflect.String && elemKind != reflect.Float64 {
			// Invalid DeviceAttribute
			return false
		}
		n := rVal1.Len()
		for i := 0; i < n; i++ {
			if !rVal1.Index(i).Equal(rVal2.Index(i)) {
				return false
			}
		}
	} else {
		if da.Value != e.Value {
			return false
		}
	}
```

model/device.go
```diff
@@ -78,6 +136,37 @@ func (da DeviceAttribute) Validate() error {
...
+		if da.Name == e.Name {
+			if da.Description == nil && e.Description == nil {
+				return true
+			}
+			if da.Description == nil || e.Description == nil {
+				return false
+			}
+			if *da.Description == *e.Description {
+				return true
+			}
```
alfrunes, line 164: The above check asserts that the attributes are not equal if both descriptions are empty (nil).
```suggestion
			if da.Description != nil {
				if e.Description == nil {
					return false
				} else if *da.Description != *e.Description {
					return false
				}
			}
```

alfrunes (changes requested):

model/device.go
merlin-northern: I tired that, but must have missed something, since ti failed for the primitive.A type. I think I missed the .Len() call.

model/device.go
merlin-northern: >The above check asserts that the attributes are not equal if both descriptions are empty (nil).

if the descriptions are nil the assumption is they are equal (see below for my reasoning)

```go
				if da.Description == nil && e.Description == nil {
					// at this point Value, Scope, and Name are equal, and both Descriptions are nil
					// shall I assume otherwise?
					return true
				}
				if da.Description == nil || e.Description == nil {
					// if either is nil while the other is not they are different
					return false
				}
				if *da.Description == *e.Description {
					// both not nil, we can compare and if equal the attributes are equal
					return true
				}
```

model/device.go
merlin-northern: this was on purpose, since it is the less common scenario. but ok, no problem.

model/device.go
merlin-northern: nice! thank you!

merlin-northern:
@mender-test-bot start pipeline please

utils/time.go
```diff
@@ -0,0 +1,43 @@
...
+		t.Year(),
+		t.Month(),
+		t.Day(),
+		0,
+		0,
+		0,
+		0,
+		t.Location(),
+	)
+}
```
alfrunes, line 30: TIP: `time` already has this in the standard library
```go
date := t.Truncate(24 * time.Hour)
// Or, to use UTC as our reference TZ
date = t.UTC().Truncate(24 * time.Hour)
```

model/device.go
```diff
@@ -78,6 +132,48 @@ func (da DeviceAttribute) Validate() error {
...
+	rVal2 := reflect.ValueOf(e.Value)
+	if rVal1.Kind() != rVal2.Kind() {
+		return false
+	}
+	if rVal1.Kind() == reflect.Slice {
+		if rVal1.Len() != rVal2.Len() {
+			return false
+		}
+		for i := 0; i < rVal1.Len(); i++ {
+			if rVal1.Index(i) != rVal2.Index(i) {
```
alfrunes, line 149: Comparing two different `reflect.Value` will always evaluate to false. You need to use `reflect.Value.Equal` to compare the underlying values.
```suggestion
			if !rVal1.Comparable() || !rVal1.Index(i).Equal(rVal2.Index(i)) {
```

store/mongo/datastore_mongo.go
```diff
@@ -311,14 +309,55 @@ func (db *DataStoreMongo) UpsertDevicesAttributesWithRevision(
 	return db.upsertAttributes(ctx, devices, attrs, false, true, "", "")
 }
 
+func (db *DataStoreMongo) inventoryNeedsUpdate(
+	ctx context.Context,
+	id model.DeviceID,
+	newAttributes model.DeviceAttributes,
+) (rc bool) {
+	rc = false
+	device, err := db.GetDevice(ctx, id)
```
alfrunes, line 318: Could we get all the devices in one batch?

alfrunes (changes requested):
Found an issue with the way attributes are compared. This also seem to not be properly covered by unit tests(?)

On a different note, I think we could highly improve the data transfer efficiency from MongoDB if we instead of comparing all the attributes, hashed the device object in a canonical form and only compared the hashes. In any case, this could be a follow-up improvement if the data transfer overhead becomes unexpectedly high.

model/device.go
merlin-northern: I changed it, because it panicked. I will check that, we have tests cases around it.

merlin-northern:
> Found an issue with the way attributes are compared. This also seem to not be properly covered by unit tests(?)
> 
> On a different note, I think we could highly improve the data transfer efficiency from MongoDB if we instead of comparing all the attributes, hashed the device object in a canonical form and only compared the hashes. In any case, this could be a follow-up improvement if the data transfer overhead becomes unexpectedly high.

that was my initial idea, to use some sum, I used md5 actually. but it was not easy, due to the fact that there is created and updated timestamp.

utils/time.go
merlin-northern: we had this discussion :) truncate does truncate, not resetting the date fields. besides we already use it in other services :)

model/device.go
merlin-northern: wow. I missed Comparable. will you believe it? I was just about to complain about lack of comparable in go, and here it is. thanks. and sorry.

merlin-northern: > Comparing two different reflect.Value will always evaluate to false. You need to use reflect.Value.Equal to compare the underlying values.

this is only true for e.g.: `[]int{1,2,3}` and `[3]int{1,2,3}` -- but this (for supported types) I managed to get.
And I remember the Comparable now, I have seen it, it does not work like that for all the cases we are testing:
```
model/device.go:149:14: rVal1.Comparable undefined (type reflect.Value has no field or method Comparable)model/device.go:149:46: rVal1.Index(i).Equal undefined (type reflect.Value has no field or method Equal)
```

And I also checked that the `rVal1.Index(i) != rVal2.Index(i)` does fill the hole.

I honestly have no other idea how to pass all the tests we are having (I had added Equal call to all of then before I asked for the second (I think) review). Are we really using all those attributes from the units in the field?

store/mongo/datastore_mongo.go
merlin-northern: sure, as soon as I am done with what I am tackling now.

merlin-northern: @alfrunes something like that: https://github.com/mendersoftware/inventory/pull/396/commits/54643f8af17d8073a503ce48a7ff05e85f8f1e13 ? I am fixing some tests now and adding more, byt could you give me a green light on that approach?

model/device.go
alfrunes: What about just comparing the values by `reflect.DeepEqual` since that is what you're essentially doing?

merlin-northern: > What about just comparing the values by `reflect.DeepEqual` since that is what you're essentially doing?

I tried DeepEqual -- it does not really work for our case. even equal elements are not considered equal.

merlin-northern: @alfrunes we have to move on with this :) I believe it should be ok for our use-cases since 99% of people have primitive types and flat attributes. do you want me to re-visit the DeepEqual and try to come up with something?

alfrunes: Comparing `reflect.Value`s does not work. Try this test case:
```go
	assert.True(t,
		DeviceAttribute{Name: "foo", Value: []interface{}{123.0}}.Equal(
			DeviceAttribute{Name: "foo", Value: []interface{}{123.0}},
		))
```
The reason your test passes is because you're comparing the same slice reference.

merlin-northern: this is a killer :) ok I will revisit this. there has to be a better way to do it.

merlin-northern: man, I swear I was fooling around DeepEqual, and I have no idea, but now all is fine. what do you think? green light?

merlin-northern:
@mender-test-bot start pipeline --fast please

model/device.go
```diff
@@ -78,6 +133,100 @@ func (da DeviceAttribute) Validate() error {
...
+				if value1 != value2 {
+					return false
+				}
+			}
+			if rVal2.Index(i).Kind() == reflect.Float64 {
+				floatComparePrecision := "%.8f"
+				floatValue1 := rVal1.Index(i).Interface().(float64)
+				floatValue2 := rVal2.Index(i).Interface().(float64)
+				value1 := fmt.Sprintf(floatComparePrecision, floatValue1)
+				value2 := fmt.Sprintf(floatComparePrecision, floatValue2)
```
alfrunes, line 181: I still don't get why you're going through this string conversion gymnastic. We don't do any floating point arithmetic so I don't think it's necessary to consider any potential floating point rounding errors on the client.

model/device.go
```diff
@@ -78,6 +133,100 @@ func (da DeviceAttribute) Validate() error {
...
+
+func (da DeviceAttribute) Equal(e DeviceAttribute) bool {
+	if (da.Value == nil || e.Value == nil) && !(da.Value == nil && e.Value == nil) {
+		return false
+	}
+	rVal1 := reflect.ValueOf(da.Value)
+	rVal2 := reflect.ValueOf(e.Value)
+	if rVal1.Kind() != rVal2.Kind() {
+		if rVal1.Kind() == reflect.Slice && rVal2.Kind() == reflect.Array ||
+			rVal1.Kind() == reflect.Array && rVal2.Kind() == reflect.Slice {
```
alfrunes, line 199: `json.Unmarshal` always unmarshal JSON arrays to slices (`[]interface{}`). The check for `reflect.Array` is redundant.

model/device.go
```diff
@@ -78,6 +133,100 @@ func (da DeviceAttribute) Validate() error {
...
+		} else {
+			return false
+		}
+	}
+	if rVal1.Kind() == reflect.Slice || rVal1.Kind() == reflect.Array {
```
alfrunes, line 204: :point_up:

alfrunes (approved):

model/device.go
merlin-northern: yes I know, but in our tests we really have slices and arrays explicite, I made this condition really to pass all the test cases we already have.

model/device.go
merlin-northern: lets say it is just in case.

PR #401, opened 2023-08-02
fix: attributes udpate: be mindful of the subset of attributes being …

merlin-northern:
…patched.

Changelog: Title
Ticket: MEN-6643

### DIFF
--- a/api/http/api_inventory.go
+++ b/api/http/api_inventory.go
@@ -507,11 +507,22 @@ func (i *inventoryHandlers) updateDeviceAttributes(
 
 	// upsert or replace the attributes
 	if r.Method == http.MethodPatch {
+		// we only check if the inventory needs update with PATCH, we assume that calling PUT
+		// is a conscious choice of the caller, to write to the db unconditionally.
+		if !i.inventory.InventoryNeedsUpdate(
+			ctx,
+			attrs,
+			deviceID,
+			scope,
+		) {
+			w.WriteHeader(http.StatusNoContent)
+			return
+		}
 		err = i.inventory.UpsertAttributesWithUpdated(ctx, deviceID, attrs, scope, etag)
 	} else if r.Method == http.MethodPut {
 		err = i.inventory.ReplaceAttributes(ctx, deviceID, attrs, scope, etag)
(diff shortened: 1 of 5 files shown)


merlin-northern:
@mender-test-bot start pipeline --fast --sugar please

merlin-northern:
@alfrunes give me green light on this idea, and I will add tests.

alfrunes (changes requested):
This does not check if some of the attributes removed. To be honest, this implementation is starting to get really complicated to follow...
Why can't we just do a very simple implementation in the app layer for the PUT/PATCH devices APIs? Something along the lines of:
 1. Sort (stable) the attributes (by name)
 2.  Remove duplicates (preserve last entry)
 3. Fetch attributes from database
 4. Sort attributes and filter by scope == "inventory"
 5. Compare the two sets of attributes

api/http/api_inventory.go
```diff
@@ -507,11 +507,22 @@ func (i *inventoryHandlers) updateDeviceAttributes(
 
 	// upsert or replace the attributes
 	if r.Method == http.MethodPatch {
+		// we only check if the inventory needs update with PATCH, we assume that calling PUT
+		// is a conscious choice of the caller, to write to the db unconditionally.
```
merlin-northern, line 511: @alfrunes heads up here: shall we keep it only in the PATCH or also check in PUT where we replace all the attributes (I am having some doubts about introducing the check in the latter case as well) (if we want this, `InventoryNeedsUpdate` will need a small update)

merlin-northern:
> This does not check if some of the attributes removed. To be honest, this implementation is starting to get really complicated to follow... Why can't we just do a very simple implementation in the app layer for the PUT/PATCH devices APIs? Something along the lines of:
> 
> 1. Sort (stable) the attributes (by name)
> 2. Remove duplicates (preserve last entry)
> 3. Fetch attributes from database
> 4. Sort attributes and filter by scope == "inventory"
> 5. Compare the two sets of attributes

okey, my friend. what do you think now?

alfrunes:
Closing as this PR is getting stale.

PR #435, opened 2023-12-19
fix: Bound the aggregation to get filter attributes

alfrunes:


alfrunes:

kjaskiewiczz (commented):
Looks like a reasonable solution, but still, if I'm not wrong, it can break things for tenants with large fleet of devices (I know you're sorting by device id to get a good sample).
Should we at least document the behavior introduced with this change?

store/mongo/datastore_mongo.go
```diff
@@ -743,6 +743,17 @@ func (db *DataStoreMongo) GetFiltersAttributes(
 	const DbCount = "count"
 
 	cur, err := collDevs.Aggregate(ctx, []bson.M{
+		// Sample 1000 devices to get a representative sample
+		{
+			"$limit": 10000,
```
merlin-northern, line 748: could we make it into a const? (maybe one day we would like to make it configurable)

alfrunes: Done

alfrunes:
@kjaskiewiczz I updated the API specs with documentation for the new imposed limitation. I also documented the existing limitation of maximum 500 attributes.

I also dropped one of the projection stages and moved the final `$limit` before the last `$sort` stage because sorting attributes from `5,000` devices can potentially exhaust the aggregation memory limit. This has a slight impact on the behavior: previously we were returning the 500 most frequent attributes, but with this change we can no longer hold this guarantee. However, this behavior is not documented anywhere and statistically the most significant attributes, so I chose to neglect this behavior change in favor of performance.

kjaskiewiczz (approved):
looks good

merlin-northern (approved):
thank you. lgtm.

alfrunes:

merlin-northern:
> I trimmed yet another unneeded projection stage from the pipeline. I tested the aggregation pipeline on a database with > 400k device documents and the aggregation runs in ~300ms compared to 80s using the old pipeline. The estimated memory consumption is also reduced to less than 300KiB for both the `$group` and the final `$sort` stages.

nice opt! you are my hero.

PR #439, opened 2024-01-24
fix: do not return updated_ts as zero time if updated_ts is not set

kjaskiewiczz:


kjaskiewiczz:

inv/inventory.go
```diff
@@ -293,7 +293,7 @@ func (i *inventory) needsUpsert(
 	needsUpsert := true
 	if device != nil {
 		// we update the inventory attributes at least once every (calendar) day
-		if !device.UpdatedTs.IsZero() &&
+		if device.UpdatedTs != nil &&
```
merlin-northern, line 296:
```suggestion
		if device.UpdatedTs != nil && !device.UpdatedTs.IsZero() &&
```

PR #455, opened 2024-05-10
fix: store "check_in_time" attribute as ISODate instead of string

kjaskiewiczz:
Ticket: MEN-7259

### DIFF
--- a/api/http/api_inventory.go
+++ b/api/http/api_inventory.go
@@ -16,6 +16,7 @@ package http
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"strconv"
 	"strings"
@@ -86,6 +87,11 @@ const (
 
 const (
 	DefaultTimeout = time.Second * 10
+)
+
+const (
+	checkInTimeParamName  = "check_in_time"
+	checkInTimeParamScope = "system"
 )
 
 // model of device's group name response at /devices/:id/group endpoint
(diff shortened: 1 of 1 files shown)


api/http/api_inventory.go
```diff
@@ -593,8 +598,16 @@ func (i *inventoryHandlers) PatchDeviceAttributesInternalHandler(
 		u.RestErrWithLog(w, r, l, err, http.StatusBadRequest)
 		return
 	}
-	for _, a := range attrs {
+	for i, a := range attrs {
 		a.Scope = r.PathParam("scope")
+		if a.Name == checkInTimeParamName {
```
tranchitella, line 603: maybe check both scope (system) and name?

kjaskiewiczz: done

kjaskiewiczz:
@mender-test-bot start pipeline --fast

tranchitella (approved):

api/http/api_inventory.go
```diff
@@ -593,8 +599,16 @@ func (i *inventoryHandlers) PatchDeviceAttributesInternalHandler(
 		u.RestErrWithLog(w, r, l, err, http.StatusBadRequest)
 		return
 	}
-	for _, a := range attrs {
+	for i, a := range attrs {
 		a.Scope = r.PathParam("scope")
```
merlin-northern, line 603: @kjaskiewiczz this will never overwrite the element in the attrs array. I know it is not the part of this TR, but i just noticed that.

kjaskiewiczz: since you noticed, maybe you can fix it? ;)

merlin-northern: https://github.com/mendersoftware/inventory/pull/456 only for you Baby.

